PR #1, opened 2016-01-14
Added read string ability

kkoser:
This commit adds a new function called keyboard_read_str, which reads
in a string from the keyboard until a return character is encountered.
It echos all input to the console as well, so the user can see what
he or she is typing. In the future, this can be made optional.

Tested in main by echoing and during comparisons with strings.

### DIFF
--- a/src/keyboard.c
+++ b/src/keyboard.c
@@ -39,6 +39,8 @@ static char buffer[BUFFER_SIZE];
 static int buffer_read = 0;
 static int buffer_write = 0;
 
+static char str_buffer[BUFFER_SIZE];
+
 static struct list queue = {0,0};
 
 static int keyboard_scan()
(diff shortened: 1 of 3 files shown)


src/keyboard.c
```diff
@@ -131,6 +133,25 @@ char keyboard_read()
...
+	char c = keyboard_read();
+	while(c != '\n' && c != '\r' && i < BUFFER_SIZE) {
+		console_putchar(c);
+		str_buffer[i++] = c;
+		
+		c = keyboard_read();
+	}
+	
+	// Mark the end of the line
+	str_buffer[i] = '\0';
```
alexmwu, line 147: breaks array bounds when i = 256

kkoser, line 147: Fixed!

rboccabe:
const char \* for the return value?
I don't think we want to give anyone the ability to overwrite the keyboard buffer, even if once the line is read in we don't use the buffer again.

jhamilt5:
Looks good

PR #2, opened 2016-01-15
Console Colors Interface

rboccabe:
This is a simple first change that creates two new functions, console_set_fgcolor and console_set_bgcolor. This provides an interface through which the fgcolor and bgcolor structs can be updated for the command line. Finally, after main.c does it's initialization work but before it starts listening for keyboard input, the output color changes from red to white.

### DIFF
--- a/src/console.c
+++ b/src/console.c
@@ -104,6 +104,12 @@ void console_init()
 	console_putstring("\nconsole: initialized\n");
 }
 
+uint8_t console_verify_color_range(uint8_t x)
+{
+    if(x < 0) return 0;
+    if(x > 255) return 255;
+    return x;
+}
 void console_set_fgcolor(uint8_t r, uint8_t g, uint8_t b)
 {
     r = console_verify_color_range(r);
@@ -124,9 +130,4 @@ void console_set_bgcolor(uint8_t r, uint8_t g, uint8_t b)
     bgcolor.b = b;
 }
 
-uint8_t console_verify_color_range(uint8_t x)
-{
-    if(x < 0) return 0;
-    if(x > 255) return 255;
-    return x;
-}
+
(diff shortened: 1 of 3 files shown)


jhamilt5:
Looks just fine to me

src/console.c
```diff
@@ -103,3 +103,18 @@ void console_init()
 	console_reset();
 	console_putstring("\nconsole: initialized\n");
 }
+
+void console_set_fgcolor(int r, int g, int b)
```
kkoser: These should probably be unsigned ints. Having negative values doesn't make a lot of sense.

src/console.c
```diff
@@ -103,3 +103,18 @@ void console_init()
...
+{
+    fgcolor.r = r % 256;
```
kkoser: Why are we modding the values? Would it make more sense to simply round them to 256, so that setting a color of like {500,500,500} would still just be white? Otherwise this behavior becomes a bit harder to predict.

kkoser:
Can you update you fork and rebase this branch so that we can merge it easily? Thanks!

src/console.h
```diff
@@ -14,6 +14,9 @@ void console_init();
 void console_putchar( char c );
 int  console_write( int unit, const void *buffer, int nblocks, int offset );
 void console_heartbeat();
+void console_set_fgcolor(uint8_t r, uint8_t g, uint8_t b);
+void console_set_bgcolor(uint8_t r, uint8_t g, uint8_t b);
+uint8_t console_verify_color_range(uint8_t x);
```
kkoser: Does this function really need to be exposed in the header file? Seems like an internal function to me.

rboccabe: Nope. I asked last Friday about whether to expose "private" functions in the headers and was told to leave the declaration in the .h anyways.

kkoser:
+1 LGTM!

alexmwu:
I know this has been closed but do we need console_verify_color_range anymore?

PR #6, opened 2016-01-18
Update to the README

jhamilt5:
README reflects changes in Makefile to build and run .iso instead of .img


kkoser:
Should we also remove the suggested tweaks section? It doesn't seem appropriate for this project, as its no longer a starting point, but eventually a fully fledged OS.

jhamilt5:
Wasn't sure if we should. I obviously removed the part about how to make an ISO. The only advantage to keeping it in would be a starting point for outsiders that happen upon our project. But I'd be fine removing it.

kkoser:
I think removing it is best. It still exists in the base kernel repo. But we don't want people seeing this project and then thinking that those are actually like a todo list. We will develop our own list of tasks

jhamilt5:
Removed the Tweaks and Projects section.

kkoser:
LGTM!

PR #7, opened 2016-01-19
Expose graphics_bitmap

alexmwu:
Expose graphics_bitmap for use in window manager and elsewhere.


kkoser:
We didn't talk about this last night, but I'm thinking now: is it smart for us to expose this function that takes in data, or should we instead expose a wrapper that takes in file names?

alexmwu:
a file name wrapper definitely makes sense. but I also think we should expose this function if users want to manipulate images in the program and display it

kkoser:
Agreed. +1

PR #8, opened 2016-01-19
Added trig functions

kkoser:
This adds implementations for sine, cosine, and tangent, as well
as M_PI. These will be needed for drawing curves in graphics. All
values are in radians.

The implementation of these functions uses a polynomial approximation
which is described at http://lab.polygonal.de/?p=205. Using an
approximation is both easier to implement and more performant than
calculating the actual taylor series. I tested this by comparing
the sin and cos implementation to the ones provided in libc over a
range of 0-2PI. They differed by at most 0.01, and averaged ~0.005.
I think this is a reasonable range given the benefits of this approach.

### DIFF
--- a/src/math.c
+++ b/src/math.c
@@ -1,0 +1,64 @@
+/*
+Copyright (C) 2015 The University of Notre Dame
+This software is distributed under the GNU General Public License.
+See the file LICENSE for details.
+*/
+
+#include "math.h"
+
+// This approximation from http://lab.polygonal.de/?p=205
+double cos( double x ) {
+	// Round x to -PI to PI
+	if (x < -3.14159265)
+		x += 6.28318531;
+	else if (x >  3.14159265)
+		x -= 6.28318531;
+
+	// cosine is sin(X + PI/2)
+	x += 1.57079632;
+	if (x >  3.14159265)
+		x -= 6.28318531;
+
+	if (x < 0)
+		return 1.27323954 * x + 0.405284735 * x * x;
+	return 1.27323954 * x - 0.405284735 * x * x;
+}
+
+// This approximation from http://lab.polygonal.de/?p=205
+double sin( double x ) {
+	// Round x to -PI to PI
+	if (x < -3.14159265)
+		x += 6.28318531;
+	else if (x >  3.14159265)
+		x -= 6.28318531;
+
+	if (x < 0)
+		return 1.27323954 * x + 0.405284735 * x * x;
+	return 1.27323954 * x - 0.405284735 * x * x;
+}
+
+// This approximation from http://lab.polygonal.de/?p=205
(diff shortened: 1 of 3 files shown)


src/math.c
```diff
@@ -0,0 +1,64 @@
+/*
+Copyright (C) 2015 The University of Notre Dame
+This software is distributed under the GNU General Public License.
+See the file LICENSE for details.
+*/
+
+#include "math.h"
+
+// This approximation from http://lab.polygonal.de/?p=205
+double cos( double x ) {
```
alexmwu: should comment that functions assume inputs are between 0 and 2pi

src/math.h
```diff
@@ -0,0 +1,41 @@
+/*
+Copyright (C) 2015 The University of Notre Dame
+This software is distributed under the GNU General Public License.
+See the file LICENSE for details.
+*/
+
+#ifndef GRAPHICS_H
```
alexmwu: MATH_H?

src/math.c
kkoser: Fixed

src/math.h
kkoser: Fixed

src/math.c
```diff
@@ -0,0 +1,125 @@
...
+			sin = .225 * (sin *-sin - sin) + sin;
+		else
+			sin = .225 * (sin * sin - sin) + sin;
+	}
+
+	return sin;
+}
+
+// This approximation from http://lab.polygonal.de/?p=205
+double tan( double x ) {
```
alexmwu: Maybe I'm missing something, but can't tan just call sin/cos?

src/math.c
```diff
@@ -0,0 +1,125 @@
...
+// This approximation from http://lab.polygonal.de/?p=205
+double cos( double x ) {
+	double cos;
+	// Round x to -PI to PI
+	if (x < -3.14159265)
+		x += 6.28318531;
+	else if (x >  3.14159265)
+		x -= 6.28318531;
+
+	// cosine is sin(X + PI/2)
```
rboccabe: Why not just add the pi/2 first and then round x to be within -pi to pi?

src/math.c
kkoser: We definitely could. I did this so that tangent doesn't have to call two other functions, but I'm open to switching it if the community thinks that calling cos/sin is better.

rboccabe: If it doesn't affect the overall performance of tan I'd agree with Alex, and I don't see how it would.

kkoser: Changed.

src/math.c
rboccabe: No need to change given the assumption of -2pi to 2pi

PR #11, opened 2016-01-22
Refactor code to match our style guidelines

zachwaterson:
This is a fairly massive pull request, as it refactors all of Thain's code. I've confirmed that nothing breaks on my machine, but additional eyes would be appreciated to see if I missed anything.

Refactors include the removal of spaces in function declarations, moving brackets to the function declaration line, and changing tabs into 4 spaces.

Tested after each commit by building and running the OS.


src/ata.c
```diff
@@ -15,298 +15,285 @@ See the file LICENSE for details.
...
 static const int ata_base[4] = {ATA_BASE0,ATA_BASE0,ATA_BASE1,ATA_BASE1};
 static int ata_interrupt_active = 0;
 static struct list queue = {0,0};
 static struct mutex ata_mutex = MUTEX_INIT;
 
-static void ata_interrupt( int intr, int code )
-{
-	ata_interrupt_active=1;
-	process_wakeup_all(&queue);
+static void ata_interrupt(int intr, int code) {
```
alexmwu, line 82: Is this the style we are going with on function decls? That is, no spaces before and after list of parameters. I personally like this.

shuyangli, line 82: I like this style as well

jhamilt5, line 82: I agree as well!

zachwaterson:
Everything should be fixed...

src/ata.c
```diff
@@ -13,298 +13,285 @@ See the file LICENSE for details.
...
 
-	// wait for the disk to calm down
-	if(!ata_wait(id,ATA_STATUS_BSY,0)) return 0;
+    // wait for the disk to calm down
+    if (!ata_wait(id,ATA_STATUS_BSY,0)) return 0;
 
-	// get the attention of the proper disk
-	outb(flags,base+ATA_FDH);
+    // get the attention of the proper disk
+    outb(flags,base+ATA_FDH);
```
kkoser: Spaces between function params?

shuyangli:
Are we ready to merge this one yet? We're pretty much blocked by this PR now.

src/interrupt.c
```diff
@@ -34,119 +34,108 @@ static const char * exception_names[] = {
...
-
-	console_printf("interrupt: ready\n");
+void interrupt_init() {
+    int i;
+    pic_init(32, 40);
+    for (i=32; i<48; i++) {
+        interrupt_disable(i);
+        interrupt_acknowledge(i);
+    }
+    for (i=0; i<32; i++) {
```
kkoser: Spaces between operators? (i < 32) Or are we not doing that?

zachwaterson: The only reasonable way to do spaces between operators is to use `indent`. I've been playing with indent, and I can emulate most of our style guide, though it adds spaces before the first element and after the last element of lists: `{ 1, 2, 3 }`. I can push up my indent file and you can see how you like it.

zachwaterson:
Alright, someone else take a look at this. I've created an `.indent.pro` profile file that contains our style guidelines. I used it to format every C file and I've added brackets to one-line statements. You can use `indent <filename>` and it will automatically use the profile file if you are in `src`.

kkoser:
Are we requiring brackets for one line statements? I'm ok with it, but I thought we were just requiring them to be on a separate, indented line

alexmwu:
I think @rboccabe and @zachwaterson felt strongly that they should have brackets and added that to the style guide. I have no strong feelings either way.

alexmwu:
I ran the indent file on _.c/_.h files, and there were a lot of weird changes on top of the already indented files. For example, it removed empty lines in some files, and it added ^Ms to list.h/.c. list.c in particular looks very weird as newlines are removed.

The indent program also adds backup files with a ~ at the end. We should see about removing them if it doesn't hurt. At the very least, they should be added to the .gitignore.

The build also still runs and works correctly.

Indent version:

```
$ indent --version
GNU indent 2.2.11
```

alexmwu:
Here's the diff if you want to take a look.

[indent.diff.txt](https://github.com/NunyaOS/Nunya/files/102600/indent.diff.txt)

zachwaterson:
I have yet to discern why list.c gets eviscerated by indent; it's the only one that is affected in that way. There are other instances of mysterious behavior, though, as you noted.

I had hoped that we wouldn't need to run indent on files that are already formatted to our specifications, and we can instead use `indent` as needed on our additional files. However, the seemingly erratic behavior of indent is concerning...

At the least, I believe that the files as committed are formatted correctly and can be merged. I can hold off on committing `indent` for now, as it isn't foolproof. How does that sound?

kkoser:
I would almost rather just focus on getting indent to work, as it is a long term solution. We can periodically run it on the whole code base, or make it a requirement to be run before a PR will be merged, and this can help guarantee code up to style standards.

zachwaterson:
Several of our files are too brittle to be subjected to `indent` (font.c, list.c, all of the assembly files), and so we would need to make a script that would exclude them from indent. For new files being committed, running `indent` shouldn't be a problem. 

I've been looking at another tool that seems less aggressive called Artistic Style (http://astyle.sourceforge.net/astyle.html). This was the tool I used to add brackets to one-line `if` statements. We can look into using `astyle` instead if you insist on having our code base conform to what one of these style programs outputs. Of course, this is "one more thing" to work on, when we really should be moving past this issue of style by now.

alexmwu:
we can look into adding indent to the make process, with some files that are "indentable." As we find solutions for files that can't be indented, we can move them to the indentable list. We should have indent backups just in case and add .gitignores for *~ files.

shuyangli:
My opinion here is to merge the properly formatted code without committing to `indent` or `astyle` for now, so we can keep working on other code; in the future, if we can figure out a robust way to use either tool, we can make another PR to set up the infrastructure for that. Thoughts?

PR #12, opened 2016-01-22
Read from disk at block and offset

jhamilt5:
Allows use of disk read at a specified block and offset.
You can read a specified number of bytes and save it into a buffer.
Tested by ata_write() of a test string to a particular block and
then loading in 25 bytes of the string at a 5 byte offset. The
correct portion of the written string was echoed back to the console.


jhamilt5:
Accidentally sent that pull request too early. Will try again in a bit.

PR #14, opened 2016-01-22
Disk Read at Block and Offset

jhamilt5:
Allows for disk read of a specified number of bytes at a particular block and offset. 
Creates files disk.h and disk.c 
To test, try inserting the following lines after the base kernel is declared ready:

```
ata_write(0, "1234567890ABCEFGHIJKLMNOPQRSTUVWXYZ", 1, 4);
char read_buffer[25];
disk_read(read_buffer, 4, 5, 25);
read_buffer[24] = '\0';
console_printf("contents: %s\n", read_buffer);
```

You should see a partial string echoed back.

### DIFF
--- a/src/ata.c
+++ b/src/ata.c
@@ -13,8 +13,6 @@ See the file LICENSE for details.
 #include "process.h"
 #include "mutex.h"
 
-#define ATA_BLOCKSIZE 512
-
 #define ATA_IRQ0	32+14
 #define ATA_IRQ1	32+15
 #define ATA_IRQ2	32+11
@@ -389,7 +387,7 @@ void ata_init()
 
 	console_printf("ata: probing devices\n");
 
-	for(i=0;i<4;i++) {
+	for(i=0;i<1;i++) {
 		if(ata_probe(i,&nblocks,&blocksize,longname)) {
 
 			console_printf("ata unit %d: %s %d MB %s\n",i,blocksize==512 ? "ata disk" : "atapi cdrom", nblocks*blocksize/1024/1024,longname);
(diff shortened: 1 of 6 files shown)


src/ata.c
```diff
@@ -389,7 +387,7 @@ void ata_init()
 
 	console_printf("ata: probing devices\n");
 
-	for(i=0;i<4;i++) {
+	for(i=0;i<1;i++) {
```
shuyangli: What is this for? I know there were timeout warning messages during boot, but I think we should figure out why `4` was there in the first place before changing it.

src/disk.c
```diff
@@ -0,0 +1,68 @@
...
+ */
+
+#include "disk.h"
+#include "ata.h"
+#include "string.h"
+#include "console.h"
+
+#define DEFAULT_ATA_UNIT 0
+
+int disk_write(char *source, int block, int offset, int num_bytes)
```
shuyangli: I'm not really comfortable merging without this implemented...

src/disk.c
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * @brief Read in data from a particular disk block
+ * @details Read exactly num_bytes characters from a specified block and offset
+ * @param destination
+ * @param start_block_index
+ * @param offset
+ * @param num_bytes
+ * @return number of bytes read
+ */
+int disk_read(char *destination, int start_block_index, int offset, int num_bytes)
```
shuyangli, line 19: Tested; `disk_read` looks good to me.

jhamilt5:
I believe I've made the changes you pointed out @shuyangli

PR #19, opened 2016-01-28
Disk Write (and disk read cleanliness)

rboccabe:
Disk writing at a specific offset is made possible by the newly implemented and publicly interfaced disk_write. It uses ata_read to read in whole blocks into block sized buffers, overwrite the parts of that block that need to be overwritten, and uses ata_write to write the whole block back.

Additionally, as @jhamilt5 and I were coding disk write, we realized that disk_read was inefficient in using  the variable bytes_after_block_offset, so we made this parallel with the more streamlined offset calculations in disk_write.

Tested by calling disk_read and disk_write with different offsets, ensuring that wrapping around disk blocks was handled correctly.


src/disk.c
```diff
@@ -15,40 +15,63 @@ See the file LICENSE for details.
...
 
-
-int disk_read(char *destination, int start_block_index, int offset,
-              int num_bytes) {
-    int blocks_needed = ((offset + num_bytes) / ATA_BLOCKSIZE) + 1;
-    if (((offset + num_bytes) % ATA_BLOCKSIZE) == 0) {
-        blocks_needed--;
+int disk_read(char *destination, int start_block_index, int offset, int num_bytes) {
+	int blocks_needed = ((offset + num_bytes) / ATA_BLOCKSIZE) + 1;
+	if(((offset + num_bytes) % ATA_BLOCKSIZE) == 0) {
```
kkoser: in other parts of the code, we have been doing this in the inverse (only add 1 if we need to, instead of removing it if we don't). Can we change this to be the same way, both for cleanliness and consistency

src/disk.c
```diff
@@ -15,40 +15,63 @@ See the file LICENSE for details.
...
-        int bytes_to_copy =
-            ((bytes_in_destination + bytes_after_block_offset) >
-             num_bytes) ? num_bytes -
-            bytes_in_destination : bytes_after_block_offset;
-
-        memcpy(destination, block_buffer + memcpy_offset, bytes_to_copy);
-        bytes_in_destination += bytes_to_copy;
-    }
+	//Take one block at a time for memory and performance and such
+	for (blocks_read_in = 0; blocks_read_in < blocks_needed; blocks_read_in++) {
```
kkoser: redundant setting of blocks_read_in

kkoser:
Looks like this uses tabs for indentation? We should be using spaces

src/disk.c
```diff
@@ -15,40 +15,63 @@ See the file LICENSE for details.
...
+		ata_read(DEFAULT_ATA_UNIT, block_buffer, 1, start_block_index + blocks_read_in);
+		if (blocks_read_in == 0) {
+			memcpy_offset = offset;
+		}
+		else {
+			memcpy_offset = 0;
+		}
```
kkoser: use ternary for simplicity?

src/disk.c
```diff
@@ -15,40 +15,63 @@ See the file LICENSE for details.
...
+		bytes_in_destination += bytes_to_copy;
+	}
+
+	return bytes_in_destination;
+}
+
+int disk_write(char *source, int start_block_index, int offset, int num_bytes) {
+	int blocks_needed = ((offset + num_bytes) / ATA_BLOCKSIZE) + 1;
+	if (((offset + num_bytes) % ATA_BLOCKSIZE) == 0) {
+		blocks_needed--;
```
kkoser: Same comment as in disk_read

src/disk.c
```diff
@@ -15,40 +15,63 @@ See the file LICENSE for details.
...
+	int bytes_to_write;
+
+	for (blocks_read_in = 0; blocks_read_in < blocks_needed; blocks_read_in++) {
+		ata_read(DEFAULT_ATA_UNIT, block_buffer, 1, start_block_index);
+		if (blocks_read_in == 0) {
+			memcpy_offset = offset;
+		}
+		else {
+			memcpy_offset = 0;
+		}
```
kkoser: ternary?

rboccabe: addressed in upcoming code review commit.

src/disk.c
rboccabe: addressed in upcoming code review commit.

src/disk.c
rboccabe: addressed in upcoming code review commit.

src/disk.c
rboccabe: addressed in upcoming code review commit.

src/disk.c
rboccabe: addressed in upcoming code review commit.

rboccabe:
Tabs replaced with spaces. I'd started using eclipse and didn't have the bit flipped for tab expand the right way.

PR #21, opened 2016-02-01
Add TravisCI

alexmwu:
Add TravisCI Integration

Travis will only test with a .travis.yml file present.

Travis will test if PRs and the master (no other) branch makes. When the PR is merged (a push) into master, Travis will run `deploy.sh`. `deploy.sh` creates a new directory called out and copies the iso into that directory. It then creates a new git repo and commits the iso changes as the initial commit. Finally, the script force pushes the `master` branch of the the repo to the `build` branch of Nunya.

Later, we can clone the build branch on Nunya and commit the .iso, rather than creating a new branch each time. This will give a full history of the all of the .iso created by PRs.

site: https://travis-ci.org/NunyaOS/Nunya


alexmwu:
~~Still need to add a Travis site for Nunya.~~

*added

rboccabe:
Fix (or don't, if I'm misunderstanding something) and ship.

alexmwu:
so I'm pretty sure Travis uses something like Docker with a base Ubuntu install. It autoclones the git repo and runs whatever command on that repo to test for a build. So, this would require `cd`ing into `src` and `make`ing. The exit codes basically tell travis whether the build passed or failed. Since we `cd` into `src`, I `cd ..` to return to the root directory of the repo (this bit is optional, but I do it just to be back in the root directory). Plus, deploy.sh is in the root dir.

rboccabe:
This was first thing this morning and it looks like I switched the git and make utilities in my mind. We for sure need the first cd and the second can't hurt.

zachwaterson:
How are we looking on this? I'd love to add a download link to the website to download the ISO, which means we need a static link to where the latest one will be hosted. Have we switched to the single build branch for version history, or is that still a work in progress?

alexmwu:
need someone to LGTM it. it currently only has the latest build. no tags yet

kkoser:
Why are we forcing the build image to overwrite the old one instead of creating a new commit? That way git can track build images for us. Also is the url of the build image static or does it change each time? That could be a problem

alexmwu:
I just had it and wanted to push it out for now. I was thinking that we version .isos by PR and then manually tag them. The url of the build image (at least for the latest commit) would be https://github.com/NunyaOS/Nunya/blob/build/basekernel.iso?raw=true. An older one would be https://github.com/NunyaOS/Nunya/blob/[commit hash]/basekernel.iso?raw=true (I believe)

kkoser:
Is it a lot of work to change it to use versioning? I agree we should get it in soon

alexmwu:
shouldn't be too bad..I don't think. I'll try to do it tonight. can we merge this though? I want to be able to already give it an .iso

alexmwu:
Alright, so the build branch should now have history. It just uses the Merge commit message ("Merge pr # ...") in master as the commit message for the build. :+1:

kkoser:
LGTM! +1 can't wait to have it

PR #22, opened 2016-02-04
Added Windowing System

kkoser:
This adds the basic windowing system for Nunya. The windowing system
provides wrappers for graphics drawing, and enforces only drawing within
a window bounds. This is accomplished by adding a new feature to the
graphics system which is an optional restrcited range. Windows use this
to enforce clipping in a simple manner, and then clear it when finished.
It is also used to ensure we are not attempting to draw offscreen and
therefore not accessing other parts of memory.

### DIFF
--- a/src/graphics.h
+++ b/src/graphics.h
@@ -19,6 +19,32 @@ int graphics_width();
 int graphics_height();
 
 /**
+ * @brief Initializes the graphics system
+ * @details Sets up the initial drawing bounds of the system
+ * and prepares to being drawing
+ */
+void graphics_init();
+
+/**
+ * @brief Sets bounds for restricted drawing
+ * @details This will restrict all drawing to the given area of the screen.
+ * Drawing outside this area will have no effect
+ *
+ * @param x The x position of the top left corner of the bounding box
+ * @param y The y position of the top left corner of the bounding box
+ * @param width The width of the bounding box
+ * @param height The height of the bounding box
+ */
+void graphics_set_bounds(int x, int y, int width, int height);
+
+/**
+ * @brief Clears the graphics drawing restrictions
+ * @details This removes any restriction on drawing area that was added
+ * by graphics_set_bounds
+ */
+void graphics_clear_bounds();
+
+/**
  * @brief Draws a line on the display from (x1, y1) to (x2, y2)
  * @details Order of points does not matter. Points outside of display will be rounded to the closest edge
  *
@@ -44,7 +70,7 @@ void graphics_line(int x1, int y1, int x2, int y2, struct graphics_color c);
  * @param graphics_color The color to draw in
  */
 void graphics_arc(int x, int y, double r, double start_theta, double end_theta,
-				  struct graphics_color c);
+                  struct graphics_color c);
 
 /**
  * @brief Draws a circle centered at given point
--- a/src/graphics.h
+++ b/src/graphics.h
@@ -21,7 +21,7 @@ int graphics_height();
 /**
  * @brief Initializes the graphics system
  * @details Sets up the initial drawing bounds of the system
- * and prepares to being drawing
+ * and prepares to begin drawing
  */
 void graphics_init();
 
--- a/src/window.c
+++ b/src/window.c
@@ -1,0 +1,68 @@
+/*
+Copyright (C) 2015 The University of Notre Dame
+This software is distributed under the GNU General Public License.
+See the file LICENSE for details.
+*/
+
+#include "window.h"
+#include "memory.h"
+
+// This function prepares to draw in a window by setting up its boundary
+// restrictions.
+// THIS FUNCTION SHOULD ALWAYS BE CALLED BEFORE MAKING A GRAPHICS DRAW
+// CALL IN THE WINDOW SYSTEM
+static inline void window_begin_draw(struct window *w) {
+    graphics_set_bounds(w->x, w->y, w->width, w->height);
+}
+
+// This function removes the restrictions used to enforce window clipping
+// on the graphics system
+// THIS FUNCTION SHOULD ALWAYS BE CALLED IF window_begin_draw WAS CALLED
+// TO FINISH THE DRAW
+static inline void window_end_draw() {
+    graphics_clear_bounds();
+}
+
+struct window *window_create(int x, int y, int width, int height) {
+    struct window *w = kmalloc(sizeof(*w));
+
+    w->x = x;
+    w->y = y;
+    w->width = width;
+    w->height = height;
+
+    return w;
+}
+
+void window_draw_line(struct window *w, int x1, int y1, int x2, int y2, struct graphics_color color) {
+    window_begin_draw(w);
+    graphics_line(x1 + w->x, y1 + w->y, x2 + w->x, y2 + w->y, color);
+    window_end_draw();
+}
+
+void window_draw_arc(struct window *w, int x, int y, double r, double start_theta, double end_theta, struct graphics_color color) {
+    window_begin_draw(w);
+    graphics_arc(x + w->x, y + w->y, r, start_theta, end_theta, color);
+    window_end_draw();
+}
+
+void window_draw_circle(struct window *w, int x, int y, double r, struct graphics_color color) {
+    window_begin_draw(w);
+    graphics_circle(x + w->x, y + w->y, r, color);
+    window_end_draw();
+}
+
+void window_draw_char(struct window *w, int x, int y, char ch, struct graphics_color fgcolor,
+                      struct graphics_color bgcolor) {
+    window_begin_draw(w);
+    graphics_char(x + w->x, y + w->y, ch, fgcolor, bgcolor);
+    window_end_draw();
+}
+
+void window_draw_bitmap(struct window *w, int x, int y, int width, int height, uint8_t * data,
+                     struct graphics_color fgcolor,
+                     struct graphics_color bgcolor) {
+    window_begin_draw(w);
+    graphics_bitmap(x + w->x, y + w->y, width, height, data, fgcolor, bgcolor);
+    window_end_draw();
+}
--- a/src/window.c
+++ b/src/window.c
@@ -1,5 +1,5 @@
 /*
-Copyright (C) 2015 The University of Notre Dame
+Copyright (C) 2016 The University of Notre Dame
 This software is distributed under the GNU General Public License.
 See the file LICENSE for details.
 */
--- a/src/window.h
+++ b/src/window.h
@@ -1,0 +1,110 @@
+/*
+Copyright (C) 2015 The University of Notre Dame
+This software is distributed under the GNU General Public License.
+See the file LICENSE for details.
+*/
+
+#ifndef WINDOW_H
+#define WINDOW_H
+
+#include "graphics.h"
+
+struct window {
+    int x;
+    int y;
+    int width;
+    int height;
+    struct graphics_color border_color;
+    struct graphics_color background_color;
+};
+
+/**
+ * @brief Creates a window for given dimens
+ * @details Creates a window with the given location and size
+ *
+ * @param x The x position of the window, relative to the screen
+ * @param y The y position of the window, relative to the screen
+ * @param width The width of the window in pixels
+ * @param height The height of the window in pixels
+ * @return A pointer to the window
+ */
+struct window *window_create(int x, int y, int width, int height);
+
+/**
+ * @brief Draws a line in the window
+ * @details Draws a line at given position in the window
+ * It will be clipped if it extends outside the window
+ * All positions are realtive to the window
+ * @param window The window to draw in
+ * @param x1 The x position of the first point in the line
+ * @param y1 The position of the first point in the line
+ * @param x2 The x position of the second point in the line
+ * @param y2 The y position of the second point in the line
+ * @param color The color of the line
+ */
+void window_draw_line(struct window *w, int x1, int y1, int x2, int y2, struct graphics_color color);
+
+/**
+ * @brief Draws an arc in the window
+ * @details Draws an arc with given size and center inside the window
+ * Curves existing outside the window bounds will be clipped. All points
+ * relative to the window
+ *
+ * @param window The window to draw in
+ * @param x The x position of the center of the arc
+ * @param y The y position of the center of the arc
+ * @param r The radius of the arc
+ * @param start_theta The starting angle of the arc
+ * @param end_theta The ending angle of the arc
+ * @param color The color to draw in
+ */
+void window_draw_arc(struct window *w, int x, int y, double r, double start_theta, double end_theta, struct graphics_color c);
+
+/**
+ * @brief Draws a circle in the window
+ * @details Draws a circle of given size and location in the window.
+ * Curves outside the windows bounds will be clipped. All locations
+ * relative to the window
+ *
+ * @param window The window to draw in
+ * @param x The x position of the center of the window
+ * @param y The y position of the center of the window
+ * @param r The radius of the circle
+ * @param color The color to draw in
+ */
+void window_draw_circle(struct window *w, int x, int y, double r, struct graphics_color color);
+
+/**
+ * @brief Draws a character in the window
+ * @details Draws given character data in the window. Draws outside the bounds
+ * of the window will be clipped. All locations relative to the window
+ *
+ * @param window The window to draw in
+ * @param x The x position of the top left corner of the char
+ * @param y The y position of the top left corner of the char
+ * @param ch The character to draw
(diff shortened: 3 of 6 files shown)


rboccabe:
So, I feel sketchy about calling window_begin_draw and window_end_draw all the time in the windowing system. I never like having comments that are all caps and say THIS NEEDS TO BE DONE IF YOU WANT TO DO THAT because new developers to the code base might never see that comment. Although we are at a low enough level that some of these might be necessary, I don't want to be lax about it if we can avoid it. I see how every process might own a window, and so we want to be able to call window_draw_shape using that window struct, but why not have graphics restrict everything and pass every graphics function a pointer to a window or NULL, then have the graphics module do all of the clipping? Because that's too close of coupling and the graphics module would have to know about windows, which it shouldn't have to know about. In that case, if we want to change the windowing system the graphics system would have to change in response to it, which probably isn't good.

One idea would be to couple it, but do so as loosely as possible. Make it a requirement for every windowing system built on this graphics system to have a "struct window" that has integers x, y, width, height that the graphics system could use to restrict itself and then pass a pointer to that window down into graphics each time and have graphics do the restrictions. This is pretty restrictive on the windowing system.

Maybe better is to have that concept in the graphics header, a "struct graphics_restrictions" or something of that sort, which every windowing system on top of this graphics system uses (it could be a member of struct window in all such windowing systems) and passes in with a graphics_draw_this() call. This way, the graphics module can restrict itself rather than relying on the windowing system to do the restriction for it because the windowing system will pass in the struct graphics_restrictions for each call.

Yet another idea is to have the graphics module be responsible for the restrictions by calling a function that is passed in by a function pointer on a void \* struct that represents the window, and having that void \* struct also passed into the graphics function.  This allows the window structure to be entirely decoupled from the graphics system, because the function passed in by pointer will be the one which deals with the window struct, while still allowing the graphics system to enforce the drawing area restrictions. I like this because it brings the burden of security down to the lowest level, no other module would be able to be built alongside the windowing system on top of graphics without at least considering security because of the required parameters. The extra two parameters to all of the graphics calls are really ugly and perhaps inefficient on x86, sure, but this brings the security of drawing anything on the screen down to the lowest level.

The benefit I see to leaving it as it is is that we could do multiple draws on a window before window_end_draw, but the benefit here is minimized because window_end_draw is not exposed, so it is only more efficient if we continue to expand the drawing library in the windowing system to calls that have multiple graphics_XXXX() calls between window_begin_draw and window_end_draw. I'm not saying that we should expose it; not exposing certainly contains the potential problem of forgetting to call window_end_draw, because everything that would call window_begin_draw and window_end_draw would be in window.c.

I've not done any outside research on how other windowing systems are built on top of graphics systems, these are just some of my thoughts, but I think it would at least be a good discussion point before we adopt this code for good.

kkoser:
@rboccabe I want to address your ideas one by one:
1. On coupling the graphics system: We considered this approach, but were very against having the graphics system depend on the windowing system. This makes it very difficult to make changes in the future and makes things far less reusable, as you know. The solution presented here avoids any form of coupling. I really don't like the idea of having graphics need a window in order to draw, as it doesn't make a lot of sense conceptually and will involve a large expansion to the graphics library, whereas I really feel that the windowing system should be a separate module.
2. On graphics_restrictions: This sounds to me like a more restrictive version of the current implementation. I don't see a lot of advantage to it, as we would be duplicating a lot of information in memory (restrictions vs. window size) and would again be making a major refactor to the graphics system. In C, it is common to have the module depend on module vars, rather than passing the restriction info through the various levels of the graphics call stack
3. This is a delegation type approach. It does decouple the two systems, but also restricts the graphics system. Thinking about it at a high level, why does the graphics system need to ask about bounds to draw something on screen? The goal of using the system implemented here is to make it optional, so that graphics remains as versatile as possible. I would also imagine this would be a fairly inefficient implementation, as every draw call would have to go back to the restrictions function to get its info, as opposed to here where the window can wrap multiple draws inside of one restriction setting. 

When Alex and I were beginning on this system, the approach we had wanted to do was to not change the graphics library at all, and have the windowing system do bounds clipping by calculating new draw locations and restricting things as necessary. This is simple for trivial draw function such as a line, but for things like curves it becomes very complicated very quickly. We thought this was a better balance of efficiency and clear implementation.

Its definitely good to be thinking about approachability of the product. If you have any thoughts on a better way to document this, I would be interested in hearing them. In my mind, this isn't that different than restrictions we see commonly in large codebases, such as the need to manually call super, but if definitely needs to be clear.

rboccabe:
@kkoser :

I agree that the implementations of 1 and 2 should be out of the running, but I guess I don't see why, for the operating system that we are creating, graphics should should be unconcerned with "asking" about its bounds. I understand that if we expand the window drawing library we get to a point where calling multiple graphics_draw_X() calls between window_begin_draw and window_end_draw calls are made, then it becomes inefficient for us to be calling this delegate for each graphics_draw_X() call, but no such functions yet exist in the windows.c file. I guess that I'm just thinking that for the purpose of this OS, we'd want to put the security as deep into the implementation as possible, which requires changing graphics a little bit more instead of just changing the windowing module.

The current implementation is a sort of peculiar mix of security tracking and enforcement where we use the windowing system to keep track of bounds of the window, and pass the bounds of the window down to graphics as a sort of "begin security enforcement" message, after which we adjust the graphics calls from window_line(0, 0, 10, 10) to graphics_line(0+w->x, 0+w->y, ...) in the windowing system. I see that it would be terribly inefficient to calculate all the points and check the bounds for each of them before making the call to graphics_draw_X() where all of the points will be calculated again for plot_point, and so I guess the way that graphics.c was handed to us in basekernel sort of begs this implementation with plot_point not being exposed. I just still feel very peculiar with the windowing security being tracked and then sort of dictated to graphics by the windowing system and then putting the burden of enforcing security onto the graphics system.

This makes it (at least for me) tempting to either have graphics in charge of security entirely and ask for its bounds before each draw, or (here's a new idea) to expose plot_point and do the calculations in the windowing system, making it responsible for both tracking and enforcing the security aspects. The latter might seem like a bit of a risk, but is no more risky than having graphics_draw_line() exposed (which can be reduced to a single plot_point). Either option would allow us to concentrate the security in one module, and exposing plot_point is no less efficient and actually probably even less impactful to the graphics module than the current implementation. That solution does require the windowing system to know the extreme display bounds of the graphics module, but it should already know that for window allocation purposes. I don't want this to drag on, so I'll tag everyone here and we'll give it another 12 or 24 hours, but I think a couple of outside opinions can't hurt the end product.

@shuyangli @zachwaterson @alexmwu @jhamilt5

kkoser:
I'm all for hearing other thoughts, thats what reviews are for. 

On the topic of security, which kernel module enforces the windows bound does not effect the strength of the security from a user program perspective. The goal of the hierarchical system is that since all drawing will eventually be done through the kernel, the user application will not have the option to draw outside its bounds. Whether the window module or graphics module checks this will be transparent to the application. 

My concern with exposing plot pixel is that then the windowing system is then basically going around the graphics lib entirely. Alex and I discussed this a bit. It would basically involve having the windowing system provide its own, separate, implementations of the drawing functions, which is not the goal here. If we go this route, we are basically eliminating the distinction between the two.

I appreciate the generality of a delegate function, but think its a bit overkill for this issue. It involves pushing an extra param on the stack for every graphics function call, and then calling that function, and then returning to graphics execution to actually draw. Adding this overhead seems like a lot as compared to finding a good way to document the importance of the window begin and end draw calls.

zachwaterson:
As I was reading through this conversation, I arrived at the same conclusion that Kyle did in his latest comment: from a security perspective, it doesn't matter which module contains the security check, since the user-level API will hide that implementation detail away anyways.

This means that our priorities should be, in order: 
1) ensuring that the security check exists at all (which it does)
2) working to make the code as maintainable/expandable as possible (a goal for all code, to some degree). 

In that regard, I agree with Kyle's approach. Building window.h over graphics.h keeps the broad (if blunt) capabilities of graphics.h intact while beginning to build out a more secure implementation of drawing on screen which we can eventually incorporate into our user-level API. 

The only part of the discussion I haven't seen fully debated yet is the messiness of requiring window_begin_draw and window_end_draw, but it is probably too low-level to be concerned about improving the API's usability, especially if it comes at the expense of maintainability. We can work to eliminate the need for remembering to call window_begin_draw and window_end_draw as we abstract further, right?

Hopefully I haven't misunderstood anything; let me know if I did.

dthain:
This is all good discussion about how to structure things.

However, let me suggest that, at this stage of the project, it's ok to have imperfect implementations.  The windowing system isn't going to be done in one commit, and you won't know exactly what is needed until you start to write code on _top_ of the windowing system and see what works and what doesn't.

And, if you let these pull requests accumulate for too long, it becomes hard for each sub-group to accommodate the other's work.

This is all a roundabout way of recommending that you should generally accept pull requests, if they appear to be moving things forward, and don't break what is already there.

src/graphics.h
```diff
@@ -19,6 +19,32 @@ int graphics_width();
 int graphics_height();
 
 /**
+ * @brief Initializes the graphics system
+ * @details Sets up the initial drawing bounds of the system
+ * and prepares to being drawing
```
rboccabe: np: begin instead of being?

src/window.c
```diff
@@ -0,0 +1,68 @@
+/*
+Copyright (C) 2015 The University of Notre Dame
```
rboccabe: 2016

src/window.h
```diff
@@ -0,0 +1,110 @@
+/*
+Copyright (C) 2015 The University of Notre Dame
```
rboccabe: 2016

src/window.c
```diff
@@ -0,0 +1,68 @@
...
+// on the graphics system
+// THIS FUNCTION SHOULD ALWAYS BE CALLED IF window_begin_draw WAS CALLED
+// TO FINISH THE DRAW
+static inline void window_end_draw() {
+    graphics_clear_bounds();
+}
+
+struct window *window_create(int x, int y, int width, int height) {
+    struct window *w = kmalloc(sizeof(*w));
+
```
rboccabe, line 28: I know that this is harmless because graphics will never draw to it, given that graphics will never set its bounds outside of video_res, but the way the code is implies that we can create windows outside of the bounds of the screen. Just something we need to be aware of.

src/window.h
```diff
@@ -0,0 +1,110 @@
...
+ * @return A pointer to the window
+ */
+struct window *window_create(int x, int y, int width, int height);
+
+/**
+ * @brief Draws a line in the window
+ * @details Draws a line at given position in the window
+ * It will be clipped if it extends outside the window
+ * All positions are realtive to the window
+ * @param window The window to draw in
```
rboccabe: style: space between end of @details and start of @params

src/window.h
```diff
@@ -0,0 +1,110 @@
...
+ * @param x1 The x position of the first point in the line
+ * @param y1 The position of the first point in the line
```
rboccabe: np: The y position

src/window.h
```diff
@@ -0,0 +1,110 @@
...
+ * @brief Draws an arc in the window
+ * @details Draws an arc with given size and center inside the window
+ * Curves existing outside the window bounds will be clipped. All points
+ * relative to the window
+ *
+ * @param window The window to draw in
+ * @param x The x position of the center of the arc
+ * @param y The y position of the center of the arc
+ * @param r The radius of the arc
+ * @param start_theta The starting angle of the arc
```
rboccabe: These angles must be between 0 and 2PI or things might get weird, as they are passed into graphics_arc which has the 0 to 2PI restriction because it passes them directly into our math.h's cos() function which has this requirement on input angles. We can add that restriction to the comments here, or we can add that robustness to the math library.

src/window.h
```diff
@@ -0,0 +1,110 @@
...
+void window_draw_arc(struct window *w, int x, int y, double r, double start_theta, double end_theta, struct graphics_color c);
+
+/**
+ * @brief Draws a circle in the window
+ * @details Draws a circle of given size and location in the window.
+ * Curves outside the windows bounds will be clipped. All locations
+ * relative to the window
+ *
+ * @param window The window to draw in
+ * @param x The x position of the center of the window
```
rboccabe: The center of the window or of the circle?

src/window.h
```diff
@@ -0,0 +1,110 @@
...
+ * @param y The y position of the center of the window
```
rboccabe: Same as above.

src/window.h
```diff
@@ -0,0 +1,110 @@
...
+ * @param fgcolor The color to draw the character
+ * @param bgcolor The color to draw the negative space
+ */
+void window_draw_char(struct window *w, int x, int y, char ch, struct graphics_color fgcolor,
+                      struct graphics_color bgcolor);
+
+/**
+ * @brief Draws data to window
+ * @details Draws the given graphics data to the screen within the given window
+ * Dawing outside the bounds of the window will result in clipping. All locations
```
rboccabe: np: Drawing (not Dawing)

src/window.h
```diff
@@ -0,0 +1,110 @@
...
+ */
+void window_draw_line(struct window *w, int x1, int y1, int x2, int y2, struct graphics_color color);
+
+/**
+ * @brief Draws an arc in the window
+ * @details Draws an arc with given size and center inside the window
+ * Curves existing outside the window bounds will be clipped. All points
+ * relative to the window
+ *
+ * @param window The window to draw in
```
rboccabe: param name is 'w', not 'window'

src/window.h
rboccabe: param name is w, not window

src/window.h
```diff
@@ -0,0 +1,110 @@
...
+ * @param color The color to draw in
+ */
+void window_draw_circle(struct window *w, int x, int y, double r, struct graphics_color color);
+
+/**
+ * @brief Draws a character in the window
+ * @details Draws given character data in the window. Draws outside the bounds
+ * of the window will be clipped. All locations relative to the window
+ *
+ * @param window The window to draw in
```
rboccabe: param name is w not window.

src/window.h
```diff
@@ -0,0 +1,110 @@
...
+void window_draw_char(struct window *w, int x, int y, char ch, struct graphics_color fgcolor,
+                      struct graphics_color bgcolor);
+
+/**
+ * @brief Draws data to window
+ * @details Draws the given graphics data to the screen within the given window
+ * Dawing outside the bounds of the window will result in clipping. All locations
+ * relative to the window
+ *
+ * @param window The window to draw in
```
rboccabe: param name is w not window.

rboccabe:
Hey guys... after running

git checkout -b kkoser-Window master
git pull https://github.com/kkoser/Nunya.git Window
make

and starting the VM, it dies on me after only showing keyboard : ready. And VirutalBox gives me a popup:

```
A critical error has occurred while running the virtual machine and the machine execution has been stopped.
For help, please see the Community section on https://www.virtualbox.org or your support contract. Please provide the contents of the log file VBox.log and the image file VBox.png, which you can find in the /home/ryan/VirtualBox VMs/OSSD5/Logs directory, as well as a description of what you were doing when this error happened. Note that you can also access the above files by selecting Show Log from the Machine menu of the main VirtualBox window.
Press OK if you want to power off the machine or press Ignore if you want to leave it as is for debugging. Please note that debugging requires special knowledge and tools, so it is recommended to press OK now.
```

Hold off on the merge please.
@kkoser can you double check and make sure that your Window branch doesn't halt your VM?

shuyangli:
@rboccabe @kkoser That may be caused by #23; try rebasing your window branch on top of `6091a64` and see if it persists

kkoser:
@rboccabe After rebasing on #23, I see no issues. I will update the code, please try again and let me know.

src/graphics.h
kkoser: fixed

src/window.c

src/window.c
kkoser, line 28: We can create windows outside of the screen. Imagine a window that was dragged to the side and half of it is off screen. This code was designed with that in mind.

src/window.h
kkoser: fixed everywhere.

shuyangli:
aside from the design questions raised above, this code looks good to me. i think we should merge it, and we can change the api design in the future.

i haven't looked hard enough to find out the decisions made in other unix-like systems, but there are precedents with _begin() and _end() type of apis (there are a few in core graphics in os x), so it may not be a huge issue. still, we may want to look back and reevaluate if we decide this isn't the right api moving forward.

let's merge after @rboccabe verifies the breaking issue?

rboccabe:
You were right about my master being behind when I checked out his branch, good catch.

PR #23, opened 2016-02-05
Separate memory and kmalloc modules

shuyangli:
Separate memory and kmalloc modules into different files. Because the memory_page function series will be used as the basis for other functions in user land, and I'm about to write some of them, it makes sense to make the separation before the modules get too big.

Discussion: are umbrella headers useful/necessary?

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -1,4 +1,6 @@
-OBJECTS=kernelcore.o main.o console.o memory.o keyboard.o clock.o interrupt.o pic.o ata.o string.o graphics.o font.o syscall.o syscall_handler.o process.o mutex.o list.o pagetable.o rtc.o disk.o math.o
+OBJECTS=kernelcore.o main.o console.o $(MEMORY_OBJS) keyboard.o clock.o interrupt.o pic.o ata.o string.o graphics.o font.o syscall.o syscall_handler.o process.o mutex.o list.o pagetable.o rtc.o disk.o math.o
+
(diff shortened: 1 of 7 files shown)


src/interrupt.c
```diff
@@ -39,12 +39,17 @@ static void unknown_exception(int i, int code) {
 
     if (i == 14) {
         asm("mov %%cr2, %0":"=r"(vaddr));
+        // When a page fault exception is thrown, test if the vaddr is mapped
```
kkoser: These comments seem unrelated to the point of this PR

shuyangli: Do you want a separate pr for comments only?

shuyangli:
Don't merge; changes coming

src/memory.h
```diff
@@ -7,41 +7,7 @@ See the file LICENSE for details.
...
-
-/**
-* @brief Frees previously allocated memory in the kernel's portion of memory
-* @detail Takes the given pointer, and finding the appropriate page in memory, marks that region as free on that page's kmalloc_page_info.
-*
-* @param to_free The pointer to the segment of memory to be free'd. This should have been obtained from a call to kmalloc.
-* @return void
-*/
-void kfree(void* to_free);
+#include "memory_raw.h"
```
kkoser: After discussion, we vote to remove this line as memory_raw is internal and should not be exposed as part of general memory usage.

kkoser:
LGTM! +1

PR #24, opened 2016-02-05
ISO DFS

rboccabe:
DO NOT MERGE!
I repeat, DO NOT MERGE

This PR is purely for code review purposes and gives proof of ability to navigate through an .iso filesystem and print out the files/directories in a DFS starting at the root directory. Test by compiling with the makefile in etc and giving an iso file as the lone command line argument, as iso.c contains its own main function. iso.h is present for the purpose of offering an interface to kernel main after we remove the library dependencies.


etc/iso.c
```diff
@@ -0,0 +1,188 @@
...
+	char rec_date_time[7];
+	char file_flags[1];
+	char file_flags_interleaved[1];
+	uint8_t interleave_gap_size;
+	char vol_seq_num[4];
+	uint8_t len_identifier;
+	char file_identifier[2048];
+};
+
+void print_n_tabs(int n);
```
kkoser: I don't think we are doing declarations for internal functions unless necessary.

etc/iso.c
```diff
@@ -0,0 +1,188 @@
...
+
+#include "iso.h"
+#include <string.h>
+#include "kerneltypes.h"
+
+#define PVD_OFFSET 156
+#define ISO_BLOCK 2048
+
+struct directory_record {
+	uint8_t length_of_record;
```
kkoser: tabs -- plz use spaces

etc/iso.c
```diff
@@ -0,0 +1,188 @@
...
+		print_dr_name(&DR);
+		printf("\n");
+	}
+	memcpy(loc_of_parent, DR.loc_of_ext, 8);
+	print_subdirectory(fp, DR.loc_of_ext, loc_of_parent, 1);
+}
+
+void print_subdirectory(FILE *fp, char *loc_of_subdir, char *loc_of_parent, int depth) {
+	fseek(fp, ISO_BLOCK * hex_to_int(loc_of_subdir + 4, 4), SEEK_SET);	// Go to extent of subdir
+	struct directory_record DR;
```
kkoser: variable names should be lower case. Upper Case is reserved for defined constants

alexmwu:
I realize that this won't be merged in, but general syntax issues (e.g., tabs not spaces and inconsistent indenting).

rboccabe:
No need to keep this around, this branch won't ever get merged.

PR #25, opened 2016-02-05
Revert "Separate memory and kmalloc modules"

shuyangli:
Reverts NunyaOS/Nunya#23

Build passes but the machine crashes; not sure what's causing the issue. Reverting.


shuyangli:
@kkoser The problem is with process.c:58, the kmalloc call failed somehow. Can you look into it?

PR #26, opened 2016-02-05
Separate memory and kmalloc modules

shuyangli:
Separate memory and kmalloc modules into different files.

Resolved previous issue in NunyaOS/Nunya#23 (reverted in #25), where the build succeeds but the machine crashes. The problem is in [src/process.c:58](https://github.com/shuyangli/Nunya/blob/4a62e4e0525743fdea2bea72cb999e1c4405ef5b/src/process.c#L58), where previously a `kmalloc` call fails somehow.

Tested by compiling and running, and the code works fine.


kkoser:
Tested and confirmed

kkoser:
LGTM! +1

PR #27, opened 2016-02-09
DO NOT MERGE - First iso_open/read/close -- DO NOT MERGE

jhamilt5:
This is the initial implementation of iso_open/read/close. These functions need to change to use atapi-based functions to work inside of VirtualBox.


rboccabe:
Does this need to be a do not merge PR, seeing as we'll have a better version up shortly?

shuyangli:
@rboccabe you can always edit the title, but until your code is ready, please mark it as "do not merge" to prevent accidents...

rboccabe:
@shuyangli What I meant was that I don't understand why we even created a PR at this stage, just to mark it as do not merge.

shuyangli:
@rboccabe I think I agree with you, that's a question for @jhamilt5...

shuyangli:
Can you guys make the directory /etc/iso and put all iso-related stuff inside?

alexmwu:
what's the difference between this and PR 24?

kkoser:
I think the goal of putting up DNM PRs is to share the code with the group and have it reviewed? Correct me if I'm wrong. There may a better way to do this though

shuyangli:
@kkoser ideally we shouldn't put up dnm pr's. maybe we could coordinate through slack?

alexmwu:
Well, I think we would want documentation of all pertinent code reviews, ideally on Github (not on Slack for sure though). So PRs are probably the best way. We could also close the PRs that aren't ready to be merged.

rboccabe:
Closing as this was just for preliminary code review

PR #4, opened 2016-01-16
switch to genisoimage for building

zachwaterson:
This switches from the deprecated mkisofs build command to genisoimage. No changes in functionality.

PR #28, opened 2016-02-10
Cherry-pick dthain/basekernel#3 and #4

shuyangli:
Cherry-pick all changes in dthain/basekernel#3 and dthain/basekernel#4.

Tested by compiling and running, and the kernel still runs fine.


dthain:
If the patch is working, then you should see the cdrom drive detected correctly during startup in VirtualBox.

shuyangli:
@dthain Yep, I see the cd-rom being detected during startup.

kkoser:
Confirmed, I see it too. +1 Ship it!

PR #29, opened 2016-02-18
Added graphics testing framework

kkoser:
This adds a framework that will be used to do performance testing of the
graphics library, and potentially other frameworks in the future. Users
create functions that test their features, and the framework will
monitor the runtime of these functions to make sure that it is within
acceptable limits. To add a test to the test suite, simply add it to the
graphics_tests array in graphics_test.h


kkoser:
Closing this to re-do as a general testing framework instead of being graphics specific, as theres really no reason for the restriction

shuyangli:
@kkoser be careful, let's not plan on a super generic testing framework; instead let's focus on getting the main parts done first (mouse, windowing, etc)

kkoser:
@shuyangli I agree. Just looking at how this framework works, there was no reason for it to only be "graphics" as it just times functions. So I'm just going to move it into a generic testing file instead of pigeon holing it

PR #30, opened 2016-02-19
Implement simplest demand paging policy

shuyangli:
This implements the simplest demand paging page fault handler. Whenever the system runs into a page fault, our handler allocates a new page, allowing the process to continue.

There exists one problem: physical memory address 0x001a0000 in VirtualBox seems special, in that whenever we touch it, our machine halts. We'll address that in a separate pull request.

Code to test this functionality is under module_tests.h@walk_memory.


dthain:
Suggest that you add your testing codes to the kernel, and make them invokable from main() so that the user can e.g. type "testpaging" or "testgraphics" to demo those pieces...

src/pagefault.c
```diff
@@ -0,0 +1,46 @@
...
+    if (pagetable_getmap(current->pagetable, vaddr, &paddr)) {
+        // If it's mapped, then it means we can't access the vaddr.
+        // Dump the process
+        console_printf("interrupt: illegal page access at vaddr %x\n",
+                       vaddr);
+        process_dump(current);
+        process_exit(0);
+    } else {
+        // Otherwise, we know we have a legit page fault
+        // printf("interrupt: page fault at %x\n", vaddr);
```
kkoser: pelase remove commented out lines

src/pagefault.c
```diff
@@ -0,0 +1,46 @@
...
+
+        // Currently we give it as much memory as we could
+        pagetable_alloc(current->pagetable, vaddr, PAGE_SIZE,
+            // TODO(SL): figure out if these flags are correct
+            PAGE_FLAG_READWRITE | PAGE_FLAG_USER | PAGE_FLAG_ALLOC);
+
+        // If we can't allocate additional memory, kill the process
+        if (0) {
```
kkoser: always false?

src/pagetable.c
```diff
@@ -26,24 +26,38 @@ struct pageentry {
...
+    uint32_t i;
     for (i = 0; i < stop; i += PAGE_SIZE) {
         pagetable_map(p, i, i, PAGE_FLAG_KERNEL | PAGE_FLAG_READWRITE);
     }
-    stop = (unsigned)video_buffer + video_xres * video_yres * 3;
-    for (i = (unsigned)video_buffer; i <= stop; i += PAGE_SIZE) {
+
+    // SL: in VirtualBox, vram is separate from ram, and no matter how much
+    // physical memory it has, video_buffer is always 0xe0000000.
+    // TODO (SL): Ensure video buffer is mapped into superviser mode even
```
kkoser: Is there a bug in jira for this? If so we should include the bug number here, ex: NUN-25

src/pagefault.c
shuyangli: This is for the future, where we'll need to check for memory allocation policy; here we don't have those policies in place yet.

src/module_tests.h
```diff
@@ -0,0 +1,7 @@
+/*
+Copyright (C) 2015 The University of Notre Dame
+This software is distributed under the GNU General Public License.
+See the file LICENSE for details.
+*/
+
+void walk_memory();
```
alexmwu, line 13: Doxygen docs, please

kkoser:
Squash commits please before merging, you have a lot

src/module_tests.c
```diff
@@ -0,0 +1,31 @@
+/*
```
kkoser, line 1: Can we call this like memory_tests or something more specific?

shuyangli:
@kkoser each commit serves a different purpose, so it makes sense to separate them.

src/pagefault.c
```diff
@@ -0,0 +1,44 @@
...
+    } else {
+        // Otherwise, we know we have a legit page fault
+        // If we can't allocate additional memory, kill the process
+        if (0) {
+            interrupt_dump_process();
+        }
+
+        // Currently we give it as much memory as we could
+        pagetable_alloc(current->pagetable, vaddr, PAGE_SIZE,
+            // TODO(SL): figure out if these flags are correct
```
kkoser: This todo seems kinda worrying to me, in a page fault handler that will be hit a lot, I'd like to know we are using the right flags haha

src/pagefault.c
```diff
@@ -0,0 +1,44 @@
+/*
```
kkoser: Do we really need a new file only for the page fault handler? Seems like it should live in the pageable module. All other modules that implement interrupts follow that pattern

src/pagetable.h
```diff
@@ -18,17 +18,109 @@ See the file LICENSE for details.
...
+/**
+ * @brief   Get the physical address of a virtual address in a page directory
+ * @details Given a virtual address, the pagetable gets its corresponding
+ *          physical address. The lookup is successful if the vaddr is mapped
+ *          to a paddr, and the page is currently in memory.
+ *
+ * @param   p       A pointer to the page directory to be looked up
+ * @param   vaddr   Virtual address to be looked up
+ * @param   paddr   A pointer to store the looked up physical address
+ * @return  1 if the lookup is successful, and 0 otherwise
```
kkoser: Can we add some discussion here of what a 0 indicates as a return value (here I believe it means that the location is not mapped)

src/pagetable.h
```diff
@@ -18,17 +18,109 @@ See the file LICENSE for details.
...
+ *          virtual page from the given pagetable.
+ *
+ * @param   p       A pointer to the pagetable to be modified
+ * @param   vaddr   A virtual address in the page
+ */
 void pagetable_unmap(struct pagetable *p, unsigned vaddr);
+
+/**
+ * @brief   Allocate pagetables and map a given virtual address and length
+ * @details Given a virtual address, the pagetable unmaps the corresponding
```
kkoser: unmaps?

src/pagetable.h
```diff
@@ -18,17 +18,109 @@ See the file LICENSE for details.
...
 void pagetable_alloc(struct pagetable *p, unsigned vaddr, unsigned length,
                      int flags);
+
+/**
+ * @brief   Delete (free) a given pagetable
+ * @details Given a page directory, delete all its contents and free the
+ *          physical memory mapped to the virtual addresses contained in the
+ *          page directory.
+ *
+ * @param   p       A pointer to the pagetable to be deleted
```
kkoser: Description indented a lot here

src/pagetable.h
```diff
@@ -18,17 +18,109 @@ See the file LICENSE for details.
...
+ */
 void pagetable_delete(struct pagetable *p);
+
+/**
+ * @brief   Load a given pagetable
+ * @details Load a pagetable into %cr3, so when we swap processes, we can make
+ *          sure the system is using the correct pagetable.
+ *
+ * @param   p   A pointer to the pagetable to be loaded into %cr3
```
kkoser: Inconsistent indentation between param name and description

shuyangli, line 111: Can you elaborate? p is the param name

kkoser, line 111: My bad, theres no problem

src/module_tests.c
```diff
@@ -0,0 +1,31 @@
...
+Copyright (C) 2015 The University of Notre Dame
+This software is distributed under the GNU General Public License.
+See the file LICENSE for details.
+*/
+
+#include "console.h"
+#include "pagetable.h"
+#include "process.h"
+
+void walk_memory() {
```
kkoser, line 11: can you change this to be a unit test, and add it to the testing framework?

kkoser, line 11: After discussion, this test is not very quantifiable, as it depends on the length of system memory and will allocate all available memory. It is not really a "unit test"

alexmwu, line 11: Still seems like something useful to have available. Add it to the command line module?

src/module_tests.c
shuyangli, line 1: The idea of this file is for any test functions we want temporarily for demo purposes. I'm imagining that after the project has stabilized and the system mostly working, we can get rid of this file in the future in favor of actual OS features.

src/module_tests.c
shuyangli, line 11: This function tries to allocate all available memory, so when it finishes the machine will halt; doesn't sound like a good idea to make it readily accessible from the command line, unless we want a sure way to halt the machine

jhamilt5:
A lot to take in, but it looks good to me.

shuyangli:
I'm talking to Jesse and he approves, merging.

PR #32, opened 2016-02-19
ISO File System Interface

jhamilt5:
Provides an ISO File System interface with six functions:
1. iso_fopen()
2. iso_fread()
3. iso_fclose()
4. iso_dopen()
5. iso_dread()
6. iso_dclose()
These open/read/close files/directories inside of a mounted ISO file system. The intent is for a mounted file system to be attached as a secondary slave optical device inside of VirtualBox, making it ATA device 3. 

Tested by attaching an ISO file system and then opening, reading, and closing both files and directories from main.c
Please note that all path/file names are in ALL CAPS, so if the file is /path/to/file.txt on your mock file system, after genisoimage, it becomes /PATH/TO/FILE.TXT


dthain:
The filesystem is coming along nicely!

Let me make a few suggestions that will help the code to be more self contained.

Does the calling of the module need visibility into the `iso_file` and `iso_dir` structures?
If not, you can move the definition of those structures into `iso.c` so that you are free
to change them without fear of breaking outside code.

`iso_dirent` would be a more kernely name for `directory_entry`

The `directory_entry` structure appears to be peculiar to the exact layout of an ISO 9660 directory entry.  Which means that if you want to swap out the ISO filesystem for another filesystem, the layout of that structure will change.  Suggest that you move that structure inside `iso.c` and instead expose a generalized `dirent` structure that only contains the items of interest to the outside user, such as the name and type of the file.

dthain:
Also suggest that you add a little code to main.c, so that if the user types "cat /x/y/z" at the kernel prompt, you can open and read a file to completion.  Again, this isn't the real end user shell, but some scaffolding you can use to test things easily until the userland is up and running.

alexmwu:
remove `/etc` directory since it is unnecessary

kkoser:
@alexmwu Any idea whats up with the travis tests on this branch?

src/iso.c
```diff
@@ -0,0 +1,489 @@
...
+#define ISO_BLOCKSIZE 2048
+#define PVD_OFFSET 16 * ISO_BLOCKSIZE
+#define ROOT_DR_OFFSET (PVD_OFFSET) + 156
+
+#define SEEK_CUR 0
+#define SEEK_SET 1
+
+// Use these to prevent multiple fetches of the same
+// ATAPI block for consecutive, single character reads
+char global_atapi_block[ATAPI_BLOCKSIZE];
```
kkoser: These should be static to restrict their scope to this module.

src/iso.c
```diff
@@ -0,0 +1,489 @@
...
+int global_atapi_extent = -1;
+
+struct iso_point {
+    int ata_unit;
+    int cur_extent;
+    int cur_offset;
+};
+
+int get_directory_record(struct iso_point *iso_p, struct directory_record *dr);
+int hex_to_int(char *src, int len);
```
kkoser: I feel like this function should live somewhere else, since other people may want to use it... Maybe we should start discussing adding a std lib, and put a todo to move it in there later?

src/iso.c
```diff
@@ -0,0 +1,489 @@
...
+ * @details Starting at the current iso_p, reads a directory record into
+ * the argument dr, moving iso_p forward.
+ *
+ * @param iso_p A pointer to a struct iso_point which is set to the beginning of a directory record in a directory extent
+ * @param dr A pointer to a directory record to be filled with the information
+ * @return 1 on success, 0 on failure
+ */
+int get_directory_record(struct iso_point *iso_p, struct directory_record *dr) {
+    int bytes_read = 0;
+    int max_dr_size = 64;
```
kkoser: This should be moved to a #define constant

src/iso.c
```diff
@@ -0,0 +1,489 @@
...
+    char dr_dest[max_dr_size];
+    int result = iso_media_read(dr_dest, 1, max_dr_size, iso_p);
+    if(result != max_dr_size) {
+        console_printf("get directory record fails\n");
+        return 0;
+    }
+
+    memcpy(&(dr->length_of_record), dr_dest + 0, sizeof(dr->length_of_record));   //1 for bytes read
+    memcpy(&(dr->length_of_ext_record), dr_dest + 1, sizeof(dr->length_of_ext_record));
```
kkoser, line 78: Lots of magic numbers in here... Idk if its reasonable to move them all into constants, but we should at least leave a comment with a link to the page that tells what they all are.

src/iso.c
```diff
@@ -0,0 +1,489 @@
...
+    if ((flags >> 1) & 1) {
+        return 1;
+    }
+    else {
+        return 0;
+    }
+}
+
+int is_valid_record(struct directory_record *dr) {
+     if(dr->rec_date_time[3] < 0 || dr->rec_date_time[3] >= 24) {  // validate hours
```
kkoser: Can we not do this in a loop, at least for the < 0 checks?

src/iso.h
```diff
@@ -0,0 +1,103 @@
...
+struct iso_file *iso_fopen(const char *pname, int ata_unit);
+
+/**
+ * @brief Closes the file
+ * @details Calls free on the file if not null.
+ *
+ * @param file Pointer to the file to be closed
+ * @return 0 if successfully closed, -1 if not.
+ */
+int iso_fclose(struct iso_file *file);
```
kkoser, line 61: Why are we having this return 0 on success and -1 on failure, instead of 1 on success and 0 on failure? We don't have to co-op the unix conventions. Same for iso_dclose

src/iso.c
```diff
@@ -0,0 +1,489 @@
...
+int global_atapi_unit = -1;
+int global_atapi_extent = -1;
+
+struct iso_point {
+    int ata_unit;
+    int cur_extent;
+    int cur_offset;
+};
+
+int get_directory_record(struct iso_point *iso_p, struct directory_record *dr);
```
kkoser: All internal functions should also be static

alexmwu:
@kkoser no idea. it says that it passed here: https://travis-ci.org/NunyaOS/Nunya/builds/110446672

src/iso.c
```diff
@@ -0,0 +1,489 @@
...
+    if ((bytes_to_file_read + file->cur_offset) > file->data_length) {
+        bytes_to_file_read = file->data_length - file->cur_offset + 1;  // Don't actually want to read off the end of the file
+        bytes_to_disk_read = bytes_to_file_read - 1;
+        at_EOF = 1;
+    }
+    int bytes_read = iso_media_read(dest, 1, bytes_to_disk_read, iso_p);
+    if(bytes_read != bytes_to_disk_read) {
+        console_printf("file reading error\n");
+        iso_media_close(iso_p);
+        return -2;
```
kkoser: Is this intentional or a typo? Theres no mention of a -2 error code in the docs

src/iso.c
```diff
@@ -0,0 +1,489 @@
...
+ * @param iso_p Pointer to the struct iso_point to begin reading at.
+ * @return The number of elements read in, -1 on error.
+ */
+int iso_media_read(void *dest, int elem_size, int num_elem, struct iso_point *stream) {
+    int bytes_needed = num_elem * elem_size;
+    if(bytes_needed <= 0) {
+        return 0;
+    }
+
+    int to_read_from_start_of_cur_extent = stream->cur_offset + bytes_needed;
```
kkoser: style: extremely long variable name.

src/iso.c
```diff
@@ -0,0 +1,489 @@
...
+    //ensure that we need to perform the read before we do
+    if (atapi_blocks_to_read > 1 ||
+        global_atapi_extent != stream->cur_extent ||
+        global_atapi_unit != stream->ata_unit) {
+        char buffer[atapi_blocks_to_read * ATAPI_BLOCKSIZE];
+        if (!atapi_read(stream->ata_unit, buffer, atapi_blocks_to_read, stream->cur_extent)) {
+            console_printf("atapi_read is 0 in iso_media_read()\n");
+            return -1;
+        }
+        else {
```
kkoser: else should be on the previous line, please fix throughout

src/iso.c
```diff
@@ -0,0 +1,489 @@
...
+    }
+}
+
+/**
+ * @brief Moves the iso_p
+ * @details Changes the offset and extent values of the iso_point as instructed
+ *
+ * @param iso_p Pointer to the iso_point to adjust
+ * @param offset How far to move the current pointer
+ * @param whence The relative place to move the pointer from, SEEK_CUR is moving forward from current position, SEEK_SET is from the zeroth byte of the iso
```
kkoser: np: naming? Seems like mode would be better, as thats what you call it in the error below.

src/iso.h
```diff
@@ -0,0 +1,103 @@
+/*
+Copyright (C) 2016 The University of Notre Dame
+This software is distributed under the GNU General Public License.
+See the file LICENSE for details.
+*/
+
+#ifndef ISO_H_
```
kkoser: extra underscore at end

src/iso.h
```diff
@@ -0,0 +1,103 @@
...
+#define ISO_H_
+
+#include "kerneltypes.h"
```
kkoser, line 10: If we are including kernel types, do we want to use the bool type that is provided in there for function returns (where appropriate) instead of an int?

src/iso.c
jhamilt5: No, -1 is the implemented definition of EOF. -2 is reserved here for a file reading error. The positive integers and zero are reserved for the number of bytes successfully read.

kkoser: Cool, makes sense. We should update the docs then to indicate what the different error codes mean.

src/iso.c
shuyangli: stdlib for kernel only?

src/iso.c
```diff
@@ -0,0 +1,489 @@
...
+}
+
+/**
+ * @brief Checks whether flags represent a directory
+ * @details Bit compares directory record file flags to the dir flag bit
+ *
+ * @param flags The file flags from a directory record
+ * @return 1 if dir bit is high, 0 otherwise
+ */
+int is_dir(int flags) {
```
shuyangli: This looks equivalent to `return (flags & 0x10)`, only more instructions.

src/iso.c
```diff
@@ -0,0 +1,489 @@
...
+     }
+     if(dr->rec_date_time[0] < 0) {    // validate year
+         return 0;
+     }
+     return 1;
+ }
+
+//Documentation in header
+int iso_dclose(struct iso_dir *dir) {
+    if(dir) {
```
shuyangli: Style: space between if and ()

src/iso.c
```diff
@@ -0,0 +1,489 @@
...
+//Documentation in header
+int iso_fread(void *dest, int elem_size, int num_elem, struct iso_file *file) {
+    struct iso_point *iso_p = iso_media_open(file->ata_unit);
+    iso_media_seek(iso_p, ISO_BLOCKSIZE * file->extent_offset + file->cur_offset, SEEK_SET);
+
+    int at_EOF = 0;
+    int bytes_to_disk_read = elem_size * num_elem;
+    int bytes_to_file_read = bytes_to_disk_read;
+    if ((bytes_to_file_read + file->cur_offset) > file->data_length) {
+        bytes_to_file_read = file->data_length - file->cur_offset + 1;  // Don't actually want to read off the end of the file
```
shuyangli: Style: these lines are really long, let's try to keep them below 80 chars.

src/iso.c
```diff
@@ -0,0 +1,489 @@
...
+        return -1;  // Treat -1 return value as EOF
+    }
+
+    iso_media_close(iso_p);
+    file->cur_offset += bytes_to_file_read;
+    return bytes_read;
+}
+
+/**
+ * @brief Acts as an fopen() for iso disk
```
shuyangli: Didn't update documentation?

src/iso.h
jhamilt5, line 61: This matches the style of f[open | read | close] in stdio.h
We do it for continuity between libraries. A close should return NULL since the file is no longer open. Technically upon failure, the behavior is undefined, and so a check for if (iso_fclose(file) < 0) would make sense there.

src/iso.c
jhamilt5: I tried replacing it with your one-liner, but the behavior changed and was no longer valid.

src/iso.c
```diff
@@ -0,0 +1,494 @@
...
+    if (extent_num < 0) {
+        return 0;
+    }
+    struct iso_dir *to_return = kmalloc(sizeof(*to_return));
+    struct iso_point *iso_p = iso_media_open(ata_unit);
+    iso_media_seek(iso_p, ATAPI_BLOCKSIZE * extent_num, SEEK_SET);
+    struct directory_record *dr = kmalloc(sizeof(*dr));
+    get_directory_record(iso_p, dr);
+    if (!is_dir(dr->file_flags[0])) {
+        console_printf("Failed to open directory\n");
```
kkoser: Do we want to leave error messages in? This library will eventually be used through sys calls, and there won't be a console. Although stubbing out the console to a logging system once we hit user mode isn't a bad idea...

src/iso.c
```diff
@@ -0,0 +1,494 @@
...
+
+//Documentation in header
+int iso_fread(void *dest, int elem_size, int num_elem, struct iso_file *file) {
+    struct iso_point *iso_p = iso_media_open(file->ata_unit);
+    iso_media_seek(iso_p, ISO_BLOCKSIZE * file->extent_offset + file->cur_offset, SEEK_SET);
+
+    int bytes_to_disk_read = elem_size * num_elem;
+    int bytes_to_file_read = bytes_to_disk_read;
+    if ((bytes_to_file_read + file->cur_offset) > file->data_length) {
+
```
kkoser: Extra line here

src/iso.c
```diff
@@ -0,0 +1,494 @@
...
+    file->cur_offset += bytes_to_file_read;
+    return bytes_read;
+}
+
+/**
+ * @brief Acts as an fclose() for iso disk
+ * @details Frees the iso_point that was used to navigate the ISO disk.
+ *
+ * @param iso_p A pointer to the iso_point to close.
+ * @return Pointer to an allocated iso_point with members initialized to 0
```
kkoser: This function is void?

src/iso.c
```diff
@@ -0,0 +1,494 @@
...
+
+/**
+ * @brief Reads in the specified amount from the iso_p
+ * @details Reads into dest the specified amount from the iso starting
+ * at the current extent and offset of iso_p
+ *
+ * @param dest A buffer to put the read information into
+ * @param elem_size The size in bytes of each element to be read
+ * @param num_elem The number of elements to read into dest
+ * @param iso_p Pointer to the struct iso_point to begin reading at.
```
kkoser: param name: stream

src/iso.c
```diff
@@ -0,0 +1,494 @@
...
+        atapi_blocks_to_read++;
+    }
+
+    //ensure that we need to perform the read before we do
+    if (atapi_blocks_to_read > 1 ||
+        global_atapi_extent != stream->cur_extent ||
+        global_atapi_unit != stream->ata_unit) {
+        char buffer[atapi_blocks_to_read * ATAPI_BLOCKSIZE];
+        if (!atapi_read(stream->ata_unit, buffer, atapi_blocks_to_read, stream->cur_extent)) {
+            console_printf("atapi_read is 0 in iso_media_read()\n");
```
kkoser: This seems more like a debugging line than a useful error message to the user. Correct me if I'm wrong tho!

src/iso.c
```diff
@@ -0,0 +1,494 @@
...
+    }
+}
+
+/**
+ * @brief Moves the iso_point
+ * @details Changes the offset and extent values of the iso_point as instructed
+ *
+ * @param iso_p Pointer to the iso_point to adjust
+ * @param offset How far to move the current pointer
+ * @param whence The relative place to move the pointer from, SEEK_CUR is moving forward from current position, SEEK_SET is from the zeroth byte of the iso file
```
kkoser, line 354: param naming: whence seems incongruous to the values here of SEEK_CUR and SEEK_SET. Different name may be better?

src/iso.c
```diff
@@ -0,0 +1,494 @@
...
+            console_printf("Illegal iso_seek mode\n");
+            break;
+    }
+}
+
+/**
+ * @brief Lookup offset the dir/file requested
+ * @details Finds the offset of the file or directory named by
+ * the full pathname starting at the root of the iso filesystem
+ * (no relative paths allowed).
```
kkoser, line 383: Does this restriction then also apply to iso_dir, which uses this function? If so, we should make sure to document it there as well.

src/iso.c
```diff
@@ -0,0 +1,494 @@
...
+ *
+ * @param pathname The string name of file or directory to lookup (with no leading /)
```
kkoser: param name: pname

src/iso.c
```diff
@@ -0,0 +1,494 @@
...
+ * @param offset The extent number to search the
```
kkoser: incorrect number of params listed in documentation

src/iso.c
```diff
@@ -0,0 +1,494 @@
...
+}
+
+/**
+ * @brief Lookup offset the dir/file requested
+ * @details Only call iso_lookup, which in turn calls iso_recursive_lookup.
+ * This recurses on the next directory in pathname
+ * by looking at the directory record given by the current
+ * disk offset, but does not handle the case of root.
+ *
+ * @param pathname The string name of file or directory to lookup (with no leading /)
```
kkoser: param name: name

src/iso.c
```diff
@@ -0,0 +1,494 @@
...
+ * @param offset The extent number to search the
```
kkoser: Incorrect params listed in docs

src/iso.h
kkoser, line 61: I guess what I'm asking is have we considered not mirroring the unix functions? We are making our own OS, we can do what we want to. For instance here, the only reason this would fail is that the file was NULL. Can we indicate that in the documentation?

src/iso.c
```diff
@@ -0,0 +1,494 @@
...
+ */
+static bool is_dir(int flags) {
+    if ((flags >> 1) & 1) {
+        return 1;
+    } else {
+        return 0;
+    }
+}
+
+static bool is_valid_record(struct directory_record *dr) {
```
kkoser: No documentation for this function

src/iso.h
shuyangli, line 61: I guess we are not worrying about POSIX?

kkoser, line 61: Were we ever planning to? This OS is extremely different from any other ones that are out there, I would imagine there are things in POSIX that don't make any sense for a hierarchical system

alexmwu, line 61: I think we should follow POSIX whenever possible, just so it's easier to contribute

kkoser, line 61: How do you mean, easier to contribute?

alexmwu, line 61: I think that outside contributors would expect functions like fopen to follow POSIX standards. for functions that diverge in our OS philosophy, we won't do this. but for standard functions I think it benefits outsiders (and even ourselves).

kkoser, line 61: Makes sense to me, just wanted to discuss it.

rboccabe, line 61: If anyone outside of us 6 wants to add to the code base, they can compare our implementation to the well documented POSIX implementation. We should only deviate when it's a cognizant, purposeful design decision to.

src/iso.c
rboccabe: Yeah, this should cause an error message to be printed in ata.c. No need to duplicate it.

src/iso.c
rboccabe: In incoming commit.

src/iso.c
rboccabe: We can add such print lines back in later. Gone for now in incoming commit.

src/iso.c
rboccabe: In incoming commit.

src/iso.c
rboccabe: Good catch. In incoming commit.

src/iso.c
rboccabe: In incoming commit.

src/iso.c
rboccabe: In incoming commit.

src/iso.c
rboccabe, line 354: Sticking with the C man pages on this one.

src/iso.c
rboccabe, line 383: In incoming commit.

src/iso.c
rboccabe: In incoming commit.

src/iso.c
rboccabe: In incoming commit

src/iso.c
rboccabe: In incoming commit.

src/iso.c
rboccabe: In incoming commit.

src/iso.c
rboccabe: In incoming commit.

src/iso.h
rboccabe: In incoming commit.

zachwaterson:
Tested and verified using my own ISO and text file; this API works for text files. Having difficulty with compiled files, however.

kkoser:
Have we considered the recommendations dthain made above, with regards to generalizing the dir structure for future interoperability with different file system types? Seems like a good idea to me, but maybe something for a future commit?

src/iso.c
```diff
@@ -0,0 +1,535 @@
...
+#define MAX_DR_SIZE 64
+
+#define SEEK_CUR 0
+#define SEEK_SET 1
+
+// Use these to prevent multiple fetches of the same
+// ATAPI block for consecutive, single character reads
+static char global_atapi_block[ATAPI_BLOCKSIZE];
+static int global_atapi_unit = -1;
+int global_atapi_extent = -1;
```
kkoser: Is there a reason this isn't static?

src/iso.c
```diff
@@ -0,0 +1,535 @@
...
+
+    // seek to current offset of iso_p
+    iso_media_seek(iso_p, read_from->extent_offset * ATAPI_BLOCKSIZE + read_from->cur_offset, SEEK_SET);
+    if (iso_media_peek_byte(iso_p) == 0) {
+        //No more records on this extent, but are there more on next extent?
+        if (iso_p->cur_extent < final_extent_of_directory) {
+            iso_media_seek(iso_p, (iso_p->cur_extent + 1) * ISO_BLOCKSIZE, SEEK_SET);
+            read_from->cur_offset = ISO_BLOCKSIZE * (iso_p->cur_extent - read_from->extent_offset);
+        }
+        else {
```
kkoser: Else should be on the same line as the previous closing brace

src/iso.c
```diff
@@ -0,0 +1,535 @@
...
+}
+
+/**
+ * @brief Acts as an fclose() for iso disk
+ * @details Frees the iso_point that was used to navigate the ISO disk.
+ *
+ * @param iso_p A pointer to the iso_point to close.
+ */
+static void iso_media_close(struct iso_point *iso_p) {
+    kfree(iso_p);
```
kkoser, line 250: Should we be doing null checks here, like we do in other close functions (i.e. iso_fclose)

src/iso.c
```diff
@@ -0,0 +1,535 @@
...
+
+    memcpy(&loc_of_parent, dr.loc_of_ext, 8);
+    root_dr_loc = bendian_chars_to_int(loc_of_parent + 4, 4);
+
+    if (strcmp(pname, "/") == 0) {
+        iso_media_close(iso_p);
+        if (is_dir_search) {
+            return root_dr_loc;
+        }
+        else {  //Tried to open "/" as file
```
kkoser: style: else should be on previous line

src/iso.c
```diff
@@ -0,0 +1,535 @@
...
+            return -1;
+        }
+    }
+    else {
```
kkoser: style: else

src/iso.c
```diff
@@ -0,0 +1,535 @@
...
+ * @param dl Pointer to integer to fill in indicating dir/file data length
+ * @param is_dir_search 1 if we are looking up a directory, 0 if a file
+ * @return The extent offset for the pname, or -1 if not found or on error
+ */
+static long int iso_recursive_look_up (const char *pname, struct iso_point *iso_p, uint32_t *dl, bool is_dir_search) {
+    int next_is_found = 0;
+
+    //Find the location of the next slash in pname
+    int next_slash_index = 0;
+    while(pname[next_slash_index] != '/' && next_slash_index < strlen(pname)) {
```
kkoser, line 444: Can we use strtok here to split apart the different pieces of the path, instead of writing our own?

src/iso.c
```diff
@@ -0,0 +1,535 @@
...
+    while (!next_is_found) {
+        if (iso_media_peek_byte(iso_p) == 0) {
+            //This extent has no more directory records, as next byte
+            //should be the record_length of the next directory record
+            //between 34 and 64, but extra space at end of extent is 0'd out.
+            if (final_extent_of_directory == iso_p->cur_extent) {
+                //And we're on the final extent, desired path not found
+                return -1;
+            }
+            else {
```
kkoser: style: else

kkoser:
I had some minor comments, but besides that it looks good. Fix it and ship it!

zachwaterson:
Last night I discovered that, due to a bug, you can't enumerate the contents of the root directory. Jesse is aware. We should get that fixed before we merge.

shuyangli:
is there a bug tracking it?

src/iso.c
jhamilt5, line 250: No, because iso_media close doesn't return anything. iso_fclose returns a status so that we match the POSIX standard. For iso_media_close, it's internal, so the user doesn't care about the status. Either we freed it or it was never properly allocated, but either way, it's gone.

kkoser, line 250: I agree that it doesn't need a return value, I was more asking if a null check is a good idea before freeing something in general

src/iso.c
jhamilt5, line 444: No, and for a good reason. strtok() is destructive, and so we cannot use it on our const char *pname. Therefore we'd have to make a copy of it and strtok() that, which works except for when the recursive lookup actually recurses. It has to have a more narrow scope to look at to reach the base case, so it does pointer arithmetic based on the location of the first slash. We can't just pass in the destructed copy of pname because 1. It's been torn down partially and 2. It's not const.

jhamilt5:
Implemented @kkoser suggestions and the bug-fix for root directory. Merging now.

PR #33, opened 2016-02-19
Commands

rboccabe:
Adds a command line module and makes main call it on everything returned by keyboard_read_str() in the main loop. Currently implements echo.


src/cmd_line.c
```diff
@@ -0,0 +1,38 @@
...
+
+//Exposed
+void cmd_line_attempt(const char * line) {
+    char line_copy[KEYBOARD_BUFFER_SIZE];
+    strcpy(line_copy, line);
+    char *first_word = strtok(line_copy, " ");
+    const char *the_rest = line + (strlen(first_word) + 1);
+
+    //Ugly case statement alternative
+    if (strcmp("echo", first_word) == 0) {
```
kkoser, line 27: Can we make this a bit more elegant than a giant list of if/else statements? Seems like we could break this out into something like keywords and function pointers, create a static list of that, and just have this iterate over that list.

jhamilt5, line 27: While that is a little more elegant, I think it's better just to get this framework set up and in place. This is easily expandable and extremely simple to understand/implement.

jhamilt5:
LGTM!

src/cmd_line.c
alexmwu, line 27: It wouldn't be that much harder and a lot more flexible. But it's good that we have this out at least

PR #34, opened 2016-02-21
Added testing framework

kkoser:
This adds the ability to create functions to test our various modules.
Tests can be added by adding a new entry to the tests array in testing.h.
The framework will both confirm that the tests return true, indicating
that all of their checks passed, and that the test elapsed within
a given amount of time.

### DIFF
--- a/src/clock.c
+++ b/src/clock.c
@@ -76,19 +76,19 @@ void clock_init() {
 }
 
 int clock_compare(clock_t a, clock_t b) {
-    if(a.seconds > b.seconds) {
-        return -1;
+    if (a.seconds > b.seconds) {
+        return 1;
     }
-    if(a.seconds == b.seconds) {
-        if(a.millis > b.millis) {
-            return -1;
+    if (a.seconds == b.seconds) {
+        if (a.millis > b.millis) {
+            return 1;
         }
-        if(a.millis == b.millis) {
+        if (a.millis == b.millis) {
             return 0;
         }
 
-        return 1;
+        return -1;
     }
 
-    return 1;
+    return -1;
 }
--- a/src/clock.c
+++ b/src/clock.c
@@ -74,3 +74,21 @@ void clock_init() {
 
     console_printf("clock: ticking\n");
 }
+
+int clock_compare(clock_t a, clock_t b) {
+    if(a.seconds > b.seconds) {
+        return -1;
+    }
+    if(a.seconds == b.seconds) {
+        if(a.millis > b.millis) {
+            return -1;
+        }
+        if(a.millis == b.millis) {
+            return 0;
+        }
+
+        return 1;
+    }
+
+    return 1;
+}
--- a/src/clock.h
+++ b/src/clock.h
@@ -19,4 +19,15 @@ clock_t clock_read();
 clock_t clock_diff(clock_t start, clock_t stop);
 void clock_wait(uint32_t millis);
 
+/**
+ * @brief Compares two times
+ * @details This is a comparison function to the relation between two times
+ * 
+ * @param a The first time
+ * @param b The second time
+ * 
+ * @return Returns -1 if a is bigger, 0 if they are the same, and 1 if b is bigger
+ */
+int clock_compare(clock_t a, clock_t b);
+
 #endif
--- a/src/clock.h
+++ b/src/clock.h
@@ -26,7 +26,7 @@ void clock_wait(uint32_t millis);
  * @param a The first time
  * @param b The second time
  * 
- * @return Returns -1 if a is bigger, 0 if they are the same, and 1 if b is bigger
+ * @return Returns -1 if a is smaller, 0 if they are the same, and 1 if a is bigger
  */
 int clock_compare(clock_t a, clock_t b);
 
--- a/src/testing.c
+++ b/src/testing.c
@@ -1,0 +1,53 @@
+/*
+Copyright (C) 2016 The University of Notre Dame
+This software is distributed under the GNU General Public License.
+See the file LICENSE for details.
+*/
+
+#include "testing.h"
+
+#include "console.h"
+
+// A flag used to record the return value of the most recent test
+static int recent_test_result = 1;
+
+// A function to time a specific test
+// Returns the amount of time that elapsed during the test
+clock_t time_test(tester test) {
+	clock_t start = clock_read();
+	recent_test_result = (test)();
+	clock_t end = clock_read();
+	return clock_diff(start, end);
+}
+
+// A test runner
+// Returns true if a test was executed in a passing amount of time
+int run_test(struct test_runner *runner) {
+	clock_t actual_time = time_test(runner->test);
+	if(clock_compare(actual_time, runner->runtime_bound) < 0 || !recent_test_result) {
+		// The test failed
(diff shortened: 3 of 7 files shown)


src/clock.c
```diff
@@ -74,3 +74,21 @@ void clock_init() {
 
     console_printf("clock: ticking\n");
 }
+
+int clock_compare(clock_t a, clock_t b) {
+    if(a.seconds > b.seconds) {
```
shuyangli: style: space between if and parentheses; also let's use if–else-if here

src/clock.h
```diff
@@ -19,4 +19,15 @@ clock_t clock_read();
...
 void clock_wait(uint32_t millis);
 
+/**
+ * @brief Compares two times
+ * @details This is a comparison function to the relation between two times
+ * 
+ * @param a The first time
+ * @param b The second time
+ * 
+ * @return Returns -1 if a is bigger, 0 if they are the same, and 1 if b is bigger
```
shuyangli: It makes more sense to return -1 if a is _smaller_, and 1 if a is _bigger_. See [strcmp documentation](http://www.cplusplus.com/reference/cstring/strcmp/).

src/testing.h
```diff
@@ -0,0 +1,38 @@
...
+#include "clock.h"
+
+/**
+ * @brief The signature of a test function
+ * @details A graphics test is a function that represents a test of one graphics feature.
+ * It should perform draw calls to the screen that attempt to cover the different use
+ * cases of that feature.
+ * @return A value indicating whether the test was successful or failed
+ */
+typedef int (*tester)(void);
```
shuyangli: Naming: I'm not sure if "tester" is enough to show that this is in fact a function. Let's name it something like `test_func_t`?

src/testing.h
```diff
@@ -0,0 +1,38 @@
...
+
+struct test_runner {
```
shuyangli: Naming: test_runner sounds like an object that runs tests. this may be more appropriately named `test_unit`?

src/testing.c
```diff
@@ -0,0 +1,53 @@
...
+
+#include "console.h"
+
+// A flag used to record the return value of the most recent test
+static int recent_test_result = 1;
+
+// A function to time a specific test
+// Returns the amount of time that elapsed during the test
+clock_t time_test(tester test) {
+	clock_t start = clock_read();
```
shuyangli: Style: spaces, not tabs

src/testing.c
```diff
@@ -0,0 +1,53 @@
...
+	}
+}
+
+void run_all_tests() {
+	int size = sizeof(tests) / sizeof(tests[0]);
+	console_printf("Running %d tests\n", size);
+	int i;
+	int failed = 0;
+	for(i = 0; i < size; i += 1) {
+		if(!run_test(&(tests[i]))) {
```
shuyangli: Is the `&` required?

shuyangli:
Functionality looks good.

src/testing.h
```diff
@@ -0,0 +1,38 @@
...
+	const char *title;
+	tester test;
+	clock_t runtime_bound;
+	// TODO: Add type or grouping so we can run specific groups of tests
+};
+
+// The list of all tests to be run when the graphics test suite is called upon
+static struct test_runner tests[] = {
```
alexmwu: is the graphics test going to be added in a separate pr?

src/testing.c
```diff
@@ -0,0 +1,53 @@
...
+clock_t time_test(tester test) {
+	clock_t start = clock_read();
+	recent_test_result = (test)();
+	clock_t end = clock_read();
+	return clock_diff(start, end);
+}
+
+// A test runner
+// Returns true if a test was executed in a passing amount of time
+int run_test(struct test_runner *runner) {
```
alexmwu: should probably add a flag here and to run_all_tests that allows results of tests to be printed out (so can run all tests and just show if they passed or not)

src/clock.c
kkoser: style fixed. Else if is unnecessary here as they all contain returns.

src/clock.h
kkoser: Good catch, I messed that up. Fixed.

src/testing.c
kkoser: You mean like just run them, and print the overall results as opposed to the pass/fail for individual tests? So you would get like a running 10 tests, 10 tests passed? Could be useful. I think maybe a separate PR?

src/testing.h
kkoser: Fixed. We should always have the _t on typedefs

src/testing.h
kkoser: Yes, tests can be added in later PRs. I didn't include any here because I thought it was already plenty big  haha. I have tests that I have used to test this.

src/testing.c
kkoser: The rest runner needs to be passed by reference, as when we pass it by value, we get the general protection fault which I believe is a result of trying to copy the struct.

src/testing.c
```diff
@@ -0,0 +1,53 @@
...
+        console_printf("Test %s FAILED with a time of %d.%ds\n", runner->title, actual_time.seconds, actual_time.millis);
+        return 0;
+    } else {
+        console_printf("Test %s passed with a time of %d.%ds\n", runner->title, actual_time.seconds, actual_time.millis);
+        return 1;
+    }
+}
+
+void run_all_tests() {
+    int size = sizeof(tests) / sizeof(tests[0]);
```
shuyangli: Just curious; since now tests doesn't contain anything, does `tests[0]` work?

alexmwu: should be 0/sizeof(struct test_unit)

src/testing.h
```diff
@@ -0,0 +1,38 @@
...
+
+struct test_unit {
+    const char *title;
+    test_func_t test;
+    clock_t runtime_bound;
+    // TODO: NUN-14 Add type or grouping so we can run specific groups of tests
+};
+
+// The list of all tests to be run when the graphics test suite is called upon
+static struct test_unit tests[] = {
```
alexmwu: shouldn't this be in the .c file so that tests aren't exposed for other .c files? also, I thought static (not static inline) was bad for .h files since each file that includes the .h will have its own copy.

alexmwu:
also confirmed that it doesn't break anything. LGTM pending my comments above

src/testing.h
kkoser: The tests array should live in the header, as this is where people will add more tests. It is not very user friendly if they have to dig through the implementation to add a test. Users should not need to know anything about the internals to add a test.

As for the static keyword, I added it to restrict the scope to only the testing module. I'm not sure what the implications of a static global in a header are, can anyone shed some light on whether this is correct?

alexmwu: http://stackoverflow.com/questions/3837490/initializing-a-static-variable-in-header

I remember having this issue in a fund comp project actually

kkoser: Makes sense, no static it is haha

shuyangli:
Looks good to me; long term we should refactor this code so it only provides a way to run tests without supplying the tests themselves, so each module can write their own test harness.

PR #35, opened 2016-02-21
[NUN-9] Block off 0x1a0000 thru 0x1c0000

shuyangli:
VirtualBox halts when any address in this range is touched
(dereferenced), so we block it off.

Tested by allocating memory with this chunk blocked off, and the system
is able to allocate all available memory.


kkoser:
LGTM!

PR #36, opened 2016-02-23
Fix bug with slave PIC

kkoser:
This solves a bug that prevented the slave PIC from working
properly. The PIC has a mask that indicates which IRQs it should listen
to, with a 1 indicating that IRQ should be ignored. Masks were not being
generated properly for IRQs that lived on the slave (>=8).

The mous uses IRQ 12, so this is a necessary fix for mouse interrupts.
Tested using mouse interrupts.


shuyangli:
Looks good.

PR #38, opened 2016-02-25
[NUN-17] Correct kmalloc integer division

rboccabe:
There was an error in taking (size % 8) as opposed to (size + 2) % 8, which caused issues with exact multiples of 8 being allocated.
Corrects [NUN-17].
[NUN-10] might be a dupe, investigating.


jhamilt5:
LGTM! Tested off of both master branch and ISO-FS branch and it merged/fixed the issue for both.

PR #39, opened 2016-02-25
Filesystem Terminal Commands

jhamilt5:
Adds several terminal commands to the operating system -- cd, pwd, ls, and cat

Behavior mimics UNIX command line behavior for these functions. Supports both relative and absolute paths. 
Test by attaching an optical disk to your VirtualBox instance as the Secondary Slave (ATA 3). Your basekernel.iso file should be attached as the Secondary Master. After the OS boots, you may move around the filesystem as you would in UNIX.


src/cmd_line.c
```diff
@@ -8,15 +8,273 @@ See the file LICENSE for details.
...
 #include "testing.h"
+#include "iso.h"
 
 #define KEYBOARD_BUFFER_SIZE 256
+#define ROOT "/"
 
-//Privates
+char cur_path[256]; // The current working directory for the command line environment
+
+void move_up_directory(char *abs_path);
```
kkoser: Do we want these functions directly in the cmd_line module? I think it makes more sense for them to live in a filesystem_commands module, and keep cmd_line pure

jhamilt5: I asked @rboccabe the same thing. We decided ultimately these do directly relate to the command line, though I think we'd be open to discussing it tomorrow in the meeting.

jhamilt5:
Moved functions into the `fs_terminal_commands` module

src/string.h
```diff
@@ -21,4 +21,8 @@ void memcpy(void *d, const void *s, unsigned length);
 
 void printf(const char *s, ...);
 
+void to_upper(char *dest, const char *src);
```
alexmwu, line 34: Doxygen comments

src/string.c
```diff
@@ -206,3 +206,43 @@ void printf(const char *s, ...) {
...
+void to_upper(char *dest, const char *src) {
+    int i;
+    for (i = 0; i < strlen(src); i++) {
+        if (src[i] >= 97 && src[i] <= 122) {
+            dest[i] = src[i] - 32;
+        } else {
+            dest[i] = src[i];
+        }
+    }
+    dest[strlen(src)] = '\0';
```
alexmwu, line 219: is this necessary?

src/string.c
```diff
@@ -206,3 +206,43 @@ void printf(const char *s, ...) {
...
+void to_lower(char *dest, const char *src) {
+    int i;
+    for (i = 0; i < strlen(src); i++) {
+        if (src[i] >= 65 && src[i] <= 90) {
+            dest[i] = src[i] + 32;
+        } else {
+            dest[i] = src[i];
+        }
+    }
+    dest[strlen(src)] = '\0';
```
alexmwu, line 231: is this necessary?

src/iso.c
```diff
@@ -113,14 +112,8 @@ static uint32_t bendian_chars_to_int(unsigned char *src, int len) {
...
 
-/**
- * @brief Checks whether flags represent a directory
- * @details Bit compares directory record file flags to the dir flag bit
- *
- * @param flags The file flags from a directory record
- * @return 1 if dir bit is high, 0 otherwise
- */
-static bool is_dir(int flags) {
+//Documentation in header
```
alexmwu: unnecessary

src/fs_terminal_commands.h
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * @brief Echo the argument supplied on a new line
+ * @param arg_line The string to be printed out
+ */
+void cmd_line_echo(const char *arg_line);
+
+/**
+ * @brief List contents for multiple directories
+ * @details List all files and directories for the directories
+ * listed in arg_line. If no argument it given, it assumes the
```
alexmwu: is given

src/fs_terminal_commands.c
```diff
@@ -0,0 +1,244 @@
...
+}
+
+
+
+// Documentation in header
+void cmd_line_pwd() {
+    console_printf("%s\n", cur_path);
+}
+
+// Documentation in header
```
alexmwu: please remove all "documentation in header" comments

src/cmd_line.c
```diff
@@ -8,27 +8,47 @@ See the file LICENSE for details.
...
+// Documentation in header
+void cmd_line_show_prompt() {
+    console_set_fgcolor(255, 100, 100);
+    console_printf("%s > ", cur_path);
+    console_set_fgcolor(255, 255, 255);
 }
 
-//Exposed
+// Documentation in header
 void cmd_line_attempt(const char * line) {
```
alexmwu: np: *line

alexmwu: np: *line

src/string.c
jhamilt5, line 219: Yes. `dest` is an empty char buffer, so we need to null terminate it. It's also the exact same size as `src` so we know where to null terminate it.

src/string.c
jhamilt5, line 231: See last.

src/string.c
alexmwu, line 219: Ah you're right. Didn't notice there was a dest

src/fs_terminal_commands.c
```diff
@@ -0,0 +1,244 @@
...
+    }
+    char path_copy[256];
+    strcpy(path_copy, path);
+    char *next_dir;
+
+    if (path[0] == '/') {   // Absolute path
+        strcpy(abs_path, "/");
+        next_dir = strtok(path_copy + 1, "/");    // Move up one character to look like a relative path in root directory
+    } else {    // Relative path
+        strcpy(abs_path, cur_path);
```
alexmwu, line 68: buffer overflow? if cur_path + relative_path > 256 then we write past the buffer. do some checking

jhamilt5, line 68: I'll look into where we could overflow the buffer, but in this line, `cur_path` is already < 256 and is copied into an empty `abs_path` buffer

src/fs_terminal_commands.c
```diff
@@ -0,0 +1,244 @@
...
+ */
+void process_path_part(const char *dname, char *abs_path) {
+    if (strcmp(dname, ".") == 0) {
+        return;
+    } else if (strcmp(dname, "..") == 0) {     // Move up one directory
+        move_up_directory(abs_path);
+        return;
+    } else {
+        move_into_directory(dname, abs_path);
+        return;
```
alexmwu, line 173: `return`s at end of void functions are unnecessary (in other functions as well)

src/fs_terminal_commands.c
alexmwu, line 68: right but say `cur_path` is 256 and you have a relative path that is not empty. that will overwrite past the size of `abs_path` (which is generally 256, but not necessarily). it would be helpful to add a size param.

jhamilt5, line 68: I have, I'm on it now. The only time this would happen though is during a call to `move_into_directory` since that's when you append new characters to the current working directory. Otherwise you're moving upwards or staying still.

alexmwu, line 68: True, just wanna make sure this doesn't happen

src/fs_terminal_commands.c
jhamilt5, line 173: I like it for the sake of clarity in control flow. If it's glaringly painful to look at, we can remove it.

src/fs_terminal_commands.c
```diff
@@ -0,0 +1,238 @@
...
+
+
+
+void cmd_line_pwd() {
+    console_printf("%s\n", cur_path);
+}
+
+void set_cur_path(const char *dpath) {
+    strcpy(cur_path, dpath);
+}
```
alexmwu, line 259: Can we do a bounds check here too?

src/cmd_line.c
```diff
@@ -8,27 +8,44 @@ See the file LICENSE for details.
 #include "console.h"
 #include "string.h"
 #include "testing.h"
+#include "iso.h"
```
kkoser: unnecessary include?

src/fs_terminal_commands.h
```diff
@@ -0,0 +1,58 @@
...
+*/
+
+#ifndef FS_TERMINAL_COMMANDS_H
+#define FS_TERMINAL_COMMANDS_H
+
+#define ROOT "/"
+
+#define BUFFER_SIZE 256
+
+char cur_path[BUFFER_SIZE]; // The current working directory for the command line environment
```
kkoser: This should probably be in the .c, not the header? We don't want people to modify it themselves

jhamilt5: I disagree. It's more visible and clear here, and if people decide they want more of a buffer, they can change it here.

shuyangli: I think it makes sense to hide the cur_path buffer in `.c`

PR #40, opened 2016-02-26
Fix division to mod

alexmwu:
Fix bug in #38


jhamilt5:
LGTM. Checked it on both my master branch and Terminal Commands branch. Both worked just fine. Shipping.

PR #41, opened 2016-03-01
[NUN-13] Marked pages as used in kmalloc

kkoser:
This changes kmalloc to get pages directly from the memory_raw module,
as opposed to the process pagetable. This allow us to mark the page
as used in the memory module, so that others do not try to use it and
it is not allocated to other areas.

Tested by creating a window, which uses kmalloc, and it still works.


shuyangli:
looks good

PR #42, opened 2016-03-02
[NUN-22] Add interrupt flag

kkoser:
This adds a flag to track the last interrupt that happened. This is
useful for crashes, and it has been added to the process_dump for this
reason.

Tested by dumping the current process and checking the last interrupt,
which was a keyboard interrupt


alexmwu:
you should probably ignore the system timer (and maybe RTC?). other than that, LGTM

PR #44, opened 2016-03-02
[NUN-21] Help Function

jhamilt5:
Add the function 'help' to the terminal and a --help flag for each
terminal command.
Tested by printing the help for each command as well as the list of all
commands from 'help'

### DIFF
--- a/src/cmd_line.c
+++ b/src/cmd_line.c
@@ -4,13 +4,15 @@ This software is distributed under the GNU General Public License.
 See the file LICENSE for details.
 */
 
-
+#include "cmd_line.h"
 #include "console.h"
 #include "string.h"
 #include "testing.h"
 #include "fs_terminal_commands.h"
 
 #define KEYBOARD_BUFFER_SIZE 256
+
+void print_all_functions();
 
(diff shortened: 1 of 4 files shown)


alexmwu:
you never add --help to test (in run_all_tests). other than that, tested and LGTM

src/cmd_line.c
```diff
@@ -56,3 +60,23 @@ void cmd_line_attempt(const char *line) {
     memset(line_copy, '\0', KEYBOARD_BUFFER_SIZE);
     return;
 }
+
+void cmd_line_help(const char *args) {
```
kkoser, line 69: Why do we need args for this?

kkoser:
Needs a rebase. Just the one comment. Fix it and ship it!

kkoser:
Can we get this fixed and merged?

src/cmd_line.c
jhamilt5, line 69: Just to follow the style of each terminal function. Gives us flexibility in case we want to add flag handlers for a call to "help"

PR #45, opened 2016-03-03
Fix PROCESS_ENTRY_POINT and PROCESS_STACK_INIT

shuyangli:
Currently they're 0x0800 0000 and 0x0fff fff0, but they should be
0x8000 0000 and 0xffff fff0.


kkoser:
Looks good to me

PR #46, opened 2016-03-17
Add command to enter user mode

shuyangli:
This contains a simple user program which calls a test system call in an infinite loop, and a console command `runproc` that, when invoked, launches the user program in user mode.

Tested by compiling and running the command, and the user process launches as expected.

### DIFF
--- a/src/cmd_line.c
+++ b/src/cmd_line.c
@@ -4,11 +4,16 @@ This software is distributed under the GNU General Public License.
 See the file LICENSE for details.
 */
 
-
+#include "cmd_line.h"
 #include "console.h"
 #include "string.h"
 #include "testing.h"
 #include "fs_terminal_commands.h"
+
+#include "iso.h"            // iso
+#include "memorylayout.h"   // PROCESS_ENTRY_POINT
+#include "process.h"        // current, pagetable_getmap
+#include "kernelcore.h"     // halt
 
 #define KEYBOARD_BUFFER_SIZE 256
 
@@ -45,6 +50,8 @@ void cmd_line_attempt(const char *line) {
         cmd_line_ls(the_rest);
     } else if (strcmp("cat", first_word) == 0) {
         cmd_line_cat(the_rest);
+    } else if (strcmp("runproc", first_word) == 0) {
+        cmd_line_run_userproc();
     }
     /*else if () {
      *...
@@ -55,4 +62,32 @@ void cmd_line_attempt(const char *line) {
     }
     memset(line_copy, '\0', KEYBOARD_BUFFER_SIZE);
     return;
+}
+
+void cmd_line_run_userproc() {
+    // Load process data
+    struct iso_dir *root_dir = iso_dopen("/", 3);
+    struct iso_file *proc_file = iso_fopen("/USERPROC.", root_dir->ata_unit);
+    uint8_t *process_data = kmalloc(proc_file->data_length);
+    int num_read = iso_fread(process_data, proc_file->data_length, 1, proc_file);
+
+    // Create a new process(page, page)
+    struct process *new_proc = process_create(PAGE_SIZE, PAGE_SIZE);
+
+    // Load the code into the proper page
+    uint32_t real_addr;
+    if (!pagetable_getmap(new_proc->pagetable, PROCESS_ENTRY_POINT, &real_addr)) {
+        console_printf("Unable to get physical address of 0x80000000\n");
+        halt();
+    }
+    // Copy data
+    memcpy((void *)real_addr, (void *)process_data, proc_file->data_length);
+    kfree(process_data);
+
+    // Push the new process onto the ready list
+    __process_set_initial_process_ready(new_proc);
+
+    // Exit the current process, enter user mode
+    console_printf("Enter user mode\n");
+    process_exit(0);
 }
--- a/src/cmd_line.h
+++ b/src/cmd_line.h
@@ -28,4 +28,10 @@ void cmd_line_init();
  */
 void cmd_line_show_prompt();
 
+/**
+ * @brief Run a designated test user process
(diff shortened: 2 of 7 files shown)


src/main.c
```diff
@@ -53,6 +53,7 @@ int kernel_main() {
     //change text color to white after bootup
     console_set_fgcolor(255, 255, 255);
 
+    // If process correctly exits, these aren't reached
```
rboccabe: remove pls

rboccabe:
Should we really be changing the makefile when not everyone has the same setup? Although it makes sense to let everyone reproduce your changes, not everyone has a ./test_iso directory. As a result of this, should cmd_line.c change to allow to run a user process? This feels like a lot to add to the codebase for just a test process in user mode.

On the other hand, changing the makefile for the ability to have a standard way to add exe files to a standard testing iso file doesn't particularly hurt, so this might be a good addition to the working version of the OS.

shuyangli:
@rboccabe I included `userproc` and `test_iso.iso` because to test launching a user process (which is the premise of this pull request), we need to somehow load it into the VM. I'd be okay with keeping only changes to actual kernel code (which isn't a lot), but I think it makes sense to include the test program to show how to run a program (for our own sake).

zachwaterson:
This is in line with what I had to do to get the process onto the ready queue, and the internal double-underscored method seems like a fine way to proceed for now, until we better decide how we want to handle queues.

src/process.c
```diff
@@ -22,7 +22,7 @@ struct list ready_list = { 0, 0 };
 void process_init() {
     // Create a dummy process with no code and no data, and load its pagetable
     // Even though it's dummy, at least kernel memory is direct mapped, so
-    // kernel code can run as usual
+    // kernel code can run as usual (specifically to enable ata interrupts)
```
kkoser, line 27: What is this about? We don't use interrupts for ATA currently.

src/process.c
```diff
@@ -180,3 +180,7 @@ void process_dump(struct process *p) {
     console_printf("esp: %x\n", s->esp);
     console_printf("eip: %x\n", s->eip);
 }
+
+void __process_set_initial_process_ready(struct process *p) {
```
kkoser, line 187: why are we switching to the __ naming convention here? We haven't used it anywhere else

src/test_userproc.c
```diff
@@ -0,0 +1,15 @@
+int main();
+int _start() {
+    return main();
+
+    // invoke system call that kills the process
```
kkoser, line 5: Spurious comment?

src/cmd_line.h
```diff
@@ -28,4 +28,10 @@ void cmd_line_init();
  */
 void cmd_line_show_prompt();
 
+/**
+ * @brief Run a designated test user process
```
kkoser, line 32: np: by saying a designated user process, it sounds like you can give a program location to this function, when in fact it is always the same user program

src/cmd_line.c
```diff
@@ -56,3 +63,31 @@ void cmd_line_attempt(const char *line) {
...
+    int num_read = iso_fread(process_data, proc_file->data_length, 1, proc_file);
+
+    // Create a new process(page, page)
+    struct process *new_proc = process_create(PAGE_SIZE, PAGE_SIZE);
+
+    // Load the code into the proper page
+    uint32_t real_addr;
+    if (!pagetable_getmap(new_proc->pagetable, PROCESS_ENTRY_POINT, &real_addr)) {
+        console_printf("Unable to get physical address of 0x80000000\n");
+        halt();
```
kkoser, line 86: do we really need to halt the whole machine here? Couldn't we just return?

src/test_userproc.c
shuyangli, line 5: When the system is properly set up, _start() should call a system call that kills the process in kernel mode. Right now we don't have it ready yet, so this is just a marker so we can come back to it in the future.

src/process.c
shuyangli, line 187: This is just a "private" function that needs to be exposed; I don't have any preference for a naming convention for this

src/process.c
shuyangli, line 27: This is per dthain

src/cmd_line.c
```diff
@@ -56,3 +63,31 @@ void cmd_line_attempt(const char *line) {
...
+    }
+    // Copy data
+    memcpy((void *)real_addr, (void *)process_data, proc_file->data_length);
+    kfree(process_data);
+
+    // Push the new process onto the ready list
+    __process_set_initial_process_ready(new_proc);
+
+    // Exit the current process, enter user mode
+    console_printf("Enter user mode\n");
```
jhamilt5, line 96: Do we want to leave this print statement in? Seems more like debug info.

src/process.c
kkoser, line 27: Id still like to get some clarification here. ATA interrupts weren't possible until the mouse code and related PIC fix went in, and the filesystem code we have uses polling instead of interrupts to read data.

src/cmd_line.c
```diff
@@ -61,3 +68,31 @@ void cmd_line_attempt(const char *line) {
     memset(line_copy, '\0', KEYBOARD_BUFFER_SIZE);
     return;
 }
+
+void cmd_line_run_userproc() {
+    // Load process data
+    struct iso_dir *root_dir = iso_dopen("/", 3);
+    struct iso_file *proc_file = iso_fopen("/USERPROC.", root_dir->ata_unit);
+    uint8_t *process_data = kmalloc(proc_file->data_length);
+    int num_read = iso_fread(process_data, proc_file->data_length, 1, proc_file);
```
kkoser, line 77: unused variable here

zachwaterson:
Let's close this, since we have the full-fledged run() command coming in soon.

PR #47, opened 2016-03-17
Add Border and String drawing

kkoser:
This adds functions to draw borders of windows, and to draw strings in
windows. When a window is created, a default border color of grey
(128,128,128) is used. To change this, the window_set_border_color()
function must be used, so that the border can be redrawn in the new color.

Tested by creating a window and drawing text to it.

### DIFF
--- a/src/window.c
+++ b/src/window.c
@@ -7,6 +7,9 @@ See the file LICENSE for details.
 #include "window.h"
 #include "memory.h"
 
+#define CHARACTER_W 8
+#define CHARACTER_H 8
+
 // This function prepares to draw in a window by setting up its boundary
 // restrictions.
 // THIS FUNCTION SHOULD ALWAYS BE CALLED BEFORE MAKING A GRAPHICS DRAW
@@ -31,7 +34,21 @@ struct window *window_create(int x, int y, int width, int height) {
     w->width = width;
     w->height = height;
 
+    // Draw the border
+    struct graphics_color bc = {128,128,128};
+    window_set_border_color(w, bc);
+
     return w;
+}
+
+void window_set_border_color(struct window *w, struct graphics_color border_color) {
+    w->border_color = border_color;
+
+    // Draw the new border
+    window_draw_line(w, 0, 0, 0, w->height, border_color);
+    window_draw_line(w, 0, 0, w->width, 0, border_color);
+    window_draw_line(w, 0, w->height, w->width, w->height, border_color);
+    window_draw_line(w, w->width, 0, w->width, w->height, border_color);
 }
 
 void window_draw_line(struct window *w, int x1, int y1, int x2, int y2, struct graphics_color color) {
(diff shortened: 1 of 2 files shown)


src/window.h
```diff
@@ -31,6 +31,13 @@ struct window {
 struct window *window_create(int x, int y, int width, int height);
 
 /**
+ * @brief Sets the border color of the window
+ * @details Sets the border color of the window and re-draws with the new color
+ * @param window The window whose border color to change
```
rboccabe, line 37: insert empty line in accordance with style guide

src/window.c
```diff
@@ -59,6 +76,20 @@ void window_draw_char(struct window *w, int x, int y, char ch, struct graphics_c
...
 
+void window_draw_string(struct window *w, int x, int y, char *str, struct graphics_color fgcolor,
+                        struct graphics_color bgcolor) {
+    int pos_h = 0;
+    int pos_w = 0;
+    while (*str) {
+        window_draw_char(w, x + pos_w, y + pos_h, *str, fgcolor, bgcolor);
+        pos_w += CHARACTER_W;
+        if (*str == '\n') {
+            pos_h += CHARACTER_H;
```
rboccabe, line 85: should we reset pos_w = 0 here? right now it seems like a newline just drops you down without resetting your x position for drawing the next character.

kkoser, line 85: yes we should thanks for catching. I only tested with single line text :O

src/window.c
```diff
@@ -59,6 +76,21 @@ void window_draw_char(struct window *w, int x, int y, char ch, struct graphics_c
     window_end_draw();
 }
 
+void window_draw_string(struct window *w, int x, int y, char *str, struct graphics_color fgcolor,
```
rboccabe: const char *?

src/window.h
```diff
@@ -90,6 +97,21 @@ void window_draw_char(struct window *w, int x, int y, char ch, struct graphics_c
...
+ * the bounds of the window, it will be clipped.
+ *
+ * @param window The window to draw in
+ * @param x The x position of the top left corner of the first character
+ * @param y The y position of the top right corner of the first character
+ * @param str The text to be drawn, null terminated
+ * @param graphics_color The color to draw the characters
+ * @param graphics_color The color to draw the negative space
+ */
+void window_draw_string(struct window *w, int x, int y, char *str, struct graphics_color fgcolor,
```
rboccabe: const char \* as mentioned in c file above

kkoser: fixed

src/window.c

src/window.h
kkoser, line 37: fixed

zachwaterson:
Fix the newline character display and we're good!

src/window.c
```diff
@@ -59,6 +76,23 @@ void window_draw_char(struct window *w, int x, int y, char ch, struct graphics_c
...
+                        struct graphics_color bgcolor) {
+    int pos_h = 0;
+    int pos_w = 0;
+    while (*str) {
+        if (*str == '\n') {
+            pos_h += CHARACTER_H;
+            pos_w = 0;
+            str++;
+            continue;
+        }
```
zachwaterson, line 89: Now it's fixed; LGTM! Ship!

PR #48, opened 2016-03-18
Add debug functions for dumping memory addresses

shuyangli:
As described.

These functions are used for tracing user stack when I'm debugging process loading.

### DIFF
--- a/src/debug_kernel.c
+++ b/src/debug_kernel.c
@@ -1,0 +1,26 @@
+/*
+Copyright (C) 2015 The University of Notre Dame
+This software is distributed under the GNU General Public License.
+See the file LICENSE for details.
+*/
+
+#include "debug_kernel.h"
+#include "console.h"
+
+void debug_dump_mem_from_addr(uint32_t *addr, int len) {
+#ifdef NUNYA_KDEBUG
+    int i;
+
+    console_printf("[DEBUG] Dumping stack starting from %x\n", addr);
+    console_printf("  addr   |   value\n");
+
+    if (len < 0) {
+        len = -len;
+        addr = addr - len + 1;
+    }
+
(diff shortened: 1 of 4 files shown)


zachwaterson:
These are pretty nice, and good work with the make flag!

src/debug_kernel.c
```diff
@@ -0,0 +1,57 @@
...
+        addr = addr - len + 1;
+    }
+
+    for (i = 0; i < len; ++i) {
+        printf("%x | %x\n", (addr + i), *(addr + i));
+    }
+#endif
+}
+
+void debug_dump_mem_from_reg(enum register_name_t reg, int len) {
```
shuyangli: Now that I think about it, this function actually wouldn't work...

zachwaterson: Function calls change the stack pointer...

src/debug_kernel.h
```diff
@@ -0,0 +1,36 @@
...
+    reg_esp
+};
+
+/**
+ * @brief   Dump memory content of a given length at a given address
+ * @detail  Dump memory content in hex, starting or ending at a given address
+ *          depending on the given length.
+ *
+ * @param   addr    The start or end address of memory to be dumped
+ * @param   len     The length of memory to be dumped in 4 bytes (32-bit
```
kkoser, line 28: Is there a case where we need the negative length functionality? Seems like it could lead to strange issues since the behavior is not that intuitive.

kkoser:
Besides the one comment, looks good to me. Fix it and ship it!

src/debug_kernel.h
shuyangli, line 28: Yes, we dumped the user stack that ends at 0xFFFFFFF0, so we need to go back up memory addresses.

PR #49, opened 2016-03-31
Window parent

kkoser:
This adds nested windows. When a window performs a draw, the actual location of that draw will be determined relative to the entire line of windows it is derived from. Windows will also be clipped by their parents bounds.    

This also adds a test that can be run at the command line, window_test, which tests the relative drawing and clipping of nested windows.

### DIFF
--- a/src/window.h
+++ b/src/window.h
@@ -110,8 +110,7 @@ void window_draw_char(struct window *w, int x, int y, char ch, struct graphics_c
  * @param graphics_color The color to draw the characters
  * @param graphics_color The color to draw the negative space
  */
-
-void window_draw_string(struct window *w, int x, int y, char *str, struct graphics_color fgcolor,
+void window_draw_string(struct window *w, int x, int y, const char *str, struct graphics_color fgcolor,
                         struct graphics_color bgcolor);
 
 /**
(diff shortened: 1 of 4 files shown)


zachwaterson:
Tested by running the code on my machine and adding a child's child window with a line inside. LGTM! 
screen shot 2016-03-31 at 11 19 51 am

zachwaterson:
If no one objects, let's ship it!

dthain:
Glad that it's coming together!  Could you add just a little decoration, so that it is recognizable as a window to the casual user?

src/window.h
```diff
@@ -109,8 +110,8 @@ void window_draw_char(struct window *w, int x, int y, char ch, struct graphics_c
  * @param graphics_color The color to draw the characters
  * @param graphics_color The color to draw the negative space
  */
-void window_draw_string(struct window *w, int x, int y, const char *str, struct graphics_color fgcolor,
-						struct graphics_color bgcolor);
+void window_draw_string(struct window *w, int x, int y, char *str, struct graphics_color fgcolor,
```
alexmwu: why do we remove the const?

kkoser: My bad, will fix

alexmwu:
LGTM!

PR #51, opened 2016-04-05
Add Window System Calls

kkoser:
This adds a preliminary set of window system calls. All drawing is done
in the current process' window. If there is no window for the calling
process, then no draw calls will occur and an error will be returned.

Tested by creating a process based on the runproc example and drawing

### DIFF
--- a/src/process.h
+++ b/src/process.h
@@ -28,6 +28,7 @@ struct process {
     char *kstack_top;
     char *stack_ptr;
     uint32_t entry;
+    struct window *window;
 };
 
 void process_init();
--- a/src/sys_window.h
+++ b/src/sys_window.h
@@ -1,0 +1,128 @@
+/*
+Copyright (C) 2016 The University of Notre Dame
+This software is distributed under the GNU General Public License.
+See the file LICENSE for details.
+*/
+
+#ifndef SYS_WINDOW_H
+#define SYS_WINDOW_H
+
+#include "sys_window_struct.h"
+
+/**
+ * @brief A struct to hold information needed to draw an arc
+ * @details This struct holds data that can be passed to draw_arc
+ * to indicate the radius and start/end angles. This information
+ * is collected here as alone, draw_arc would need more parameters
+ * than are available to system calls
+ */
+struct arc_info {
+	double r;
+	double start_theta;
+	double end_theta;
+};
+
+// Window System Calls
+
+/**
+ * @brief Creates a window for the process
+ * @details Invokes the window_create function with given parameters and assigns
+ * the resulting window to calling process
+ * 
+ * @param x The x position of the window, relative to its parent
+ * @param y The y position of the window, relative to its parent
+ * @param width The width of the window
+ * @param height The height of the window
+ * @return 0 if a window was created, -1 if the process already has a widnow, otherwise error
+ */
+static inline int32_t create_window(int x, int y, uint32_t width, uint32_t height) {
+	return syscall(SYSCALL_window_create, x, y, width, height, 0);
+}
+
+/**
+ * @brief Sets the border color of the process' window
+ * @details Invokes the window_set_border_color function to change window color
+ * 
+ * @param graphics_color The color to set the border to
+ * @return 0 on success, or -1 if the calling process has no window
+ */
+static inline int32_t set_border_color(const struct graphics_color *border_color) {
+	return syscall(SYSCALL_window_set_border_color, (int)border_color, 0, 0, 0, 0);
+}
+
+/**
+ * @brief Draws a line in the current process' window
+ * @details Invokes the window_draw_line function to draw into the current window
+ * 
+ * @param x1 The x position of the first point in the line
+ * @param y1 The y position of the first point in the line
+ * @param x2 The x position of the second point in the line
+ * @param y2 The y position of the second point in the line
+ * @param color The color of the line
+ * @return 0 on success, or -1 if the calling process has no window
+ */
+static inline int32_t draw_line(int x1, int y1, int x2, int y2, const struct graphics_color *color) {
+	return syscall(SYSCALL_window_draw_line, x1, y1, x2, y2, (int)color);
+}
+
+/**
+ * @brief Draws an arc in the current process' window
+ * @details Invokes the window_draw_arc function to draw into the current window
+ * 
+ * @param x The x position of the center of the arc
+ * @param y The y position of the center of the arc
+ * @param arc The arc_info describing the arc length and radius
+ * @param color The color to draw in
+ * @return 0 on success, or -1 if the calling process has no window
+ */
+static inline int32_t draw_arc(int x, int y, const struct arc_info *arc, const struct graphics_color *c) {
+	return syscall(SYSCALL_window_draw_arc, x, y, (int)arc, (int)c, 0);
+}
+
+/**
+ * @brief Draws a circle in the current process' window
+ * @details Invokes the window_draw_circle function to draw into the current window
+ * 
+ * @param x The x position of the center of the circle, relative to the window
+ * @param y The y position of the center of the circle, relative to the window
+ * @param r The radius of the circle
+ * @param c The color to draw in
+ * @return 0 on success, or -1 if the calling process has no window
+ */
+static inline int32_t draw_circle(int x, int y, const double *r, const struct graphics_color *c) {
+	return syscall(SYSCALL_window_draw_circle, x, y, (int)r, (int)c, 0);
+}
+
+/**
+ * @brief Draws a character in the current process' window
+ * @details Invokes the window_draw_char function to draw into the current window
+ * 
+ * @param x The x position of the top left corner of the char
+ * @param y The y position of the top left corner of the char
+ * @param c The character to draw
+ * @param fgcolor The color to draw the character
+ * @param bgcolor The color to draw the negative space
+ * @return 0 on success, or -1 if the calling process has no window
+ */
+static inline int32_t draw_char(int x, int y, char c, const struct graphics_color *fgcolor, const struct graphics_color *bgcolor) {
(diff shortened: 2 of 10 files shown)


src/process.h
```diff
@@ -28,6 +28,7 @@ struct process {
     char *kstack_top;
     char *stack_ptr;
     uint32_t entry;
+    struct window *window;
```
zachwaterson, line 31: This will have to be updated to the new location of the process struct (if merged)

src/syscall.h
```diff
@@ -8,14 +8,36 @@ See the file LICENSE for details.
...
+ * @details This struct holds data that can be passed to draw_arc
+ * to indicate the radius and start/end angles. This information
+ * is collected here as alone, draw_arc would need more parameters
+ * than are available to system calls
+ */
+struct arc_info {
+	double r;
+	double start_theta;
+	double end_theta;
+};
```
zachwaterson: This should probably go in the global_structs.h file (if merged)

src/syscall.h
```diff
@@ -8,14 +8,36 @@ See the file LICENSE for details.
 #define SYSCALL_H
 
 #include "kerneltypes.h"
+#include "graphics.h"
```
zachwaterson: Is this for graphics_color? Then we should move the declaration of the struct to global_types.h as well. I think we want to minimize the amount of files syscall.h is importing.

src/syscall_handler.c
```diff
@@ -23,6 +26,49 @@ uint32_t sys_testcall(uint32_t code) {
...
+    return 0;
+}
+
+int32_t sys_draw_string(int x, int y, const char *str, const struct graphics_color *fgcolor,
+    const struct graphics_color *bgcolor) {
+    CHECK_PROC_WINDOW();
+    window_draw_string(current->window, x, y, str, *fgcolor, *bgcolor);
+    return 1;
+}
+
```
zachwaterson: I think we should move these implementations into their own file (like I did with process lifecycle) to keep the handler function quasi-manageable.

src/syscall.h
kkoser: I get that, but I'm also very wary of creating a big grab bag file that has tons of unrelated structs. Id rather keep them together with the code that needs them and have sys call.h include multiple files than have every file than needs any struct including ALL of them unnecessarily.

rboccabe: I'm on board with Kyle here.
One crazy large, perhaps disorganized or difficult to keep organized file of random structs doesn't sound too appealing. Or at the very least instead of global_structs we should call it sys_structs.h or something to give an idea of what belongs in the file.

src/syscall_handler.c
rboccabe: As a result, the .h documentation on these current implementations in syscall.h which will become one or two liners can be moved to that windows_sys.h file or whatever you call it, and the documentation in syscall.h should point to those functions.

src/syscall.h
```diff
@@ -28,4 +50,108 @@ static inline int32_t yield() {
...
+ * @details Invokes the window_create function with given parameters and assigns
+ * the resulting window to calling process
+ * 
+ * @param x The x position of the window, relative to its parent
+ * @param y The y position of the window, relative to its parent
+ * @param width The width of the window
+ * @param height The height of the window
+ * @return 0 if a window was created, otherwise error
+ */
+static inline int32_t create_window(int x, int y, int width, int height) {
```
rboccabe: should these params be of type uintX_t to avoid negative arguments?

src/syscall.h
shuyangli: At the same time we want to make sure users when they include `syscall.h` don't get a ton of random structs either. Can we make a header file within the graphics module that contains all structs exposed to the public?

src/syscall.h
```diff
@@ -8,14 +8,36 @@ See the file LICENSE for details.
...
 #define SYSCALL_testcall 2
 #define SYSCALL_yield    3
 
+#define SYSCALL_window_create 200
+#define SYSCALL_window_set_border_color 201
+#define SYSCALL_window_draw_line 202
+#define SYSCALL_window_draw_arc 203
+#define SYSCALL_window_draw_circle 204
+#define SYSCALL_window_draw_char 205
+#define SYSCALL_window_draw_string 206
```
shuyangli, line 22: Do we need a `draw_string` if we have `draw_char`? Similarly, do we need `draw_circle` if we have `draw_arc`? There could be one more layer of abstraction.

src/syscall.h
shuyangli: This shouldn't be in `syscall.h` at least

src/syscall.h
kkoser, line 22: True. The functions already exist at the kernel level (as convenience functions) so I figured it doesn't hurt to add them to sys calls. We have lots of sys call numbers haha

shuyangli, line 22: Yeah but at the same time if we really adhere to a minimalistic philosophy when we design syscalls, we should only keep the ones we need.

rboccabe, line 22: Devil's advocate for both sides here. Indeed, these could all be put into a user library which moves this one level of abstraction up and uses a bunch of calls to a draw_pixel sys_call. This keeps our syscall interface small. At the same time, that could be expensive in terms of performance if we keep entering exiting user mode and putting calls on the function stack. What would Linus do?

src/syscall.h
kkoser: It actually makes sense to allow a negative x and y -- imagine if you were creating a window which was uncentered. The sides outside of the parents bound would simply be clipped. Width and height should be positive tho, as the behavior with negative width would be undefined.

src/process.h
kkoser, line 31: Process struct not being moved.

src/syscall.h
kkoser: moved to sys_window.h, as this is a pure wrapper struct

src/syscall.h
kkoser: I am going to move graphics_color into sys_window, and then have graphics include that. It seems the simplest solution here without exposing pure graphics functions to the users

src/syscall_handler_window.h
```diff
@@ -0,0 +1,23 @@
+/*
+Copyright (C) 2016 The University of Notre Dame
+This software is distributed under the GNU General Public License.
+See the file LICENSE for details.
+*/
+
```
rboccabe, line 6: #ifndef SYSCALL_HANDLER_WINDOW_H

src/sys_window.h
```diff
@@ -0,0 +1,130 @@
...
+	return syscall(SYSCALL_window_draw_circle, x, y, (int)r, (int)c, 0);
+}
+
+/**
+ * @brief Draws a character in the current process' window
+ * @details Invokes the window_draw_char function to draw into the current window
+ * 
+ * @param x The x position of the top left corner of the char
+ * @param y The y position of the top left corner of the char
+ * @param ch The character to draw
```
rboccabe: c

src/sys_window.h
```diff
@@ -0,0 +1,130 @@
...
+}
+
+/**
+ * @brief Draws an arc in the current process' window
+ * @details Invokes the window_draw_arc function to draw into the current window
+ * 
+ * @param x The x position of the center of the arc
+ * @param y The y position of the center of the arc
+ * @param r The radius of the arc
+ * @param start_theta The starting angle of the arc, between 0 and 2*PI. See sin/cos
```
rboccabe: out of date with struct arc_info

src/sys_window.h
```diff
@@ -0,0 +1,130 @@
...
+}
+
+/**
+ * @brief Draws a circle in the current process' window
+ * @details Invokes the window_draw_circle function to draw into the current window
+ * 
+ * @param x The x position of the center of the circle, relative to the window
+ * @param y The y position of the center of the circle, relative to the window
+ * @param r The radius of the circle
+ * @param color The color to draw in
```
rboccabe: c

src/sys_window.h
```diff
@@ -0,0 +1,130 @@
...
+}
+
+/**
+ * @brief Draws a string in the current process' window
+ * @details Invokes the window_draw_string function to draw into the current window
+ * 
+ * @param x The x position of the top left corner of the first character
+ * @param y The y position of the top right corner of the first character
+ * @param str The text to be drawn, null terminated
+ * @param graphics_color The color to draw the characters
```
rboccabe: fgcolor and bgcolor

src/syscall_handler_window.c
```diff
@@ -0,0 +1,53 @@
...
+This software is distributed under the GNU General Public License.
+See the file LICENSE for details.
+*/
+#include "syscall_handler_window.h"
+#include "process.h"
+
+#define CHECK_PROC_WINDOW() if(current->window == 0) return -1
+
+int32_t sys_window_create(int x, int y, int width, int height) {
+    CHECK_PROC_WINDOW();
```
rboccabe: Why is this "called" here? You step into sys_window_create() and the first thing it seems to do is return -1 if current->window is null. Is this the intended behavior?

src/sys_window.h
kkoser: fixed

src/sys_window.h
kkoser: fixed

src/sys_window.h
kkoser: feed

src/sys_window.h
kkoser: fixed

src/syscall_handler_window.c
kkoser: Great catch, that was a little copy pasta mistake :)

src/syscall_handler_window.h
kkoser, line 6: fixed

PR #55, opened 2016-04-06
[DO NOT MERGE] Add run() syscall

zachwaterson:
This pull request accomplishes a few things:

1) Creates a global_structs.h file for those structs that need to be exposed at the user level for syscall interfacing.

2) Separates the structure of syscall implementation into smaller syscall module implementation files, which are consolidated into syscall.h

3) Implements the process lifecycle syscalls, including run(), yield(), and exit()

4) Implements a template process so others can see / test how to use the API (temporary for now)

5) Cleans up our ISO image usage and creates a dedicated `files` iso, with a `/bin/` folder for compiled programs to rest (will be used by the upcoming console program). The Makefile is having a few issues with expanding on this mechanism; stay tuned


src/cmd_line.c
```diff
@@ -49,6 +51,11 @@ void cmd_line_attempt(const char *line) {
         cmd_line_ls(the_rest);
     } else if (strcmp("cat", first_word) == 0) {
         cmd_line_cat(the_rest);
+    } else if (strcmp("process", first_word) == 0) { // temporary, for debugging
+        struct process_permissions test_permissions;
+        test_permissions.max_number_of_pages = 10;
```
kkoser, line 56: I think we need to have people mallocing the permissions struct right? Since the permissions need to persist for the entire life of the process, we can't rely on this object always being in scope. 

I feel like we should follow that convention here so we don't lead people down the wrong path

src/global_structs.h
```diff
@@ -0,0 +1,39 @@
+/*
+Copyright (C) 2015 The University of Notre Dame
+This software is distributed under the GNU General Public License.
+See the file LICENSE for details.
+*/
+
+// This file contains all structs that are needed on both the kernel level and the user level
```
kkoser, line 7: Can these be doxygen comments?

src/memory_raw.c
```diff
@@ -13,15 +13,15 @@ See the file LICENSE for details.
...
-static uint32_t pages_free = 0;
-static uint32_t pages_total = 0;
-
 static uint32_t *freemap = 0;
 static uint32_t freemap_bits = 0;
 static uint32_t freemap_bytes = 0;
 static uint32_t freemap_cells = 0;
 static uint32_t freemap_pages = 0;
 
+static uint32_t pages_free = 0;
```
kkoser, line 22: why have these been moved? We should avoid changing files unnecessarily.

src/cmd_line.c
shuyangli, line 56: I'm okay with this, since the permission struct is always copied into the process struct

src/Makefile
```diff
@@ -29,6 +31,15 @@ bootblock: bootblock.o
 kernel: ${OBJECTS}
 	ld ${KERNEL_LDFLAGS} -Ttext 0x10000 -s --oformat binary ${OBJECTS} -o $@
 
+files.iso: bin/test_userproc_even.nun #this should be bin/%.nun but it doesn't work"
```
shuyangli, line 34: let's try `$(wildcard bin/*.nun)`?

src/sys_process_lifecycle.c
```diff
@@ -0,0 +1,97 @@
...
+    return 0;
+}
+
+uint32_t sys_run(char *process_path, struct process_permissions *child_permissions) {
+    if (child_permissions == 0) {
+        return -1;
+    }
+
+    // Load process data
+    struct iso_dir *root_dir = iso_dopen("/", 3);
```
kkoser, line 29: It seems weird that we require process to be loaded from ATA devices 3. Can we not pass the device number in, or otherwise indicate which device is desired in the file path?

src/sys_process_lifecycle.c
```diff
@@ -0,0 +1,97 @@
...
+        return -1;
+    }
+
+    // store the current number of pages used, so we can see how many the child uses
+    int page_count_before_child = current->number_of_pages_using;
+
+    // Create a new process(page, page)
+    struct process *child_proc = process_create(PAGE_SIZE, PAGE_SIZE);
+    child_proc->permissions = *child_permissions; // store as value, but we needed to pass as reference because syscall
+    child_proc->parent = current; // store the child's parent
```
kkoser: Is it possible that the scheduler could switch contexts during this function call? If so, we should change this to take the desired parent as a parameter to ensure we use the right one

src/sys_process_lifecycle.c
```diff
@@ -0,0 +1,97 @@
...
+    if (!pagetable_getmap(child_proc->pagetable, PROCESS_ENTRY_POINT, &real_addr)) {
+        console_printf("Unable to get physical address of 0x80000000\n");
+        sys_exit(-1); // todo: end more cleanly
+    }
+
+    // Copy data
+    memcpy((void *)real_addr, (void *)process_data, proc_file->data_length);
+
+    // transfer pages used count to child
+    int child_pages_used = current->number_of_pages_using - page_count_before_child;
```
kkoser: Can we do all this math once and then assign? Doing all of this struct dereferencing is wasteful

src/sys_process_lifecycle.c
```diff
@@ -0,0 +1,97 @@
...
+    child_proc->number_of_pages_using += child_pages_used;
+    current->number_of_pages_using = page_count_before_child;
+
+    // add the max number of pages child could use to parent's usage
+    current->number_of_pages_using += child_proc->permissions.max_number_of_pages;
+
+    // check if we've exceeded the parent's allocation
+    if (current->number_of_pages_using > current->permissions.max_number_of_pages) {
+        console_printf("current process exceeded limit: %d > %d\n", current->number_of_pages_using, current->permissions.max_number_of_pages);
+        sys_exit(-1);
```
kkoser, line 82: Why are we using -1 for errors? I thought we had decided that error codes are returns > 0. Also we should give different returns for different errors (parent OOM and child OOM return same number)

src/sys_process_lifecycle.c
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    // check if we've exceeded the child's allocation
+    if (child_proc->number_of_pages_using > child_proc->permissions.max_number_of_pages) {
+        console_printf("child process exceeded limit\n");
+        sys_exit(-1);
+    }
+
+    // free the intermediary memory we used
+    kfree(process_data);
```
kkoser: we should free this memory immediately. Right now, this could potentially leak if an error occurs and the process exits without finishing this function

src/syscall.h
```diff
@@ -8,24 +8,93 @@ See the file LICENSE for details.
...
     return syscall(SYSCALL_yield, 0, 0, 0, 0, 0);
 }
 
+/**
+ * @brief   Creates and begins execution of a new process.
+ * @details Loads a new process into memory as a child of the current process,
+ *          and then begins execution of that process. The current process
+ *          yields and is placed at the end of the ready queue.
+ *
+ * @param   process_path The filesystem path of the process, which should be
```
kkoser: np: not the path of the process, but of the executable file

src/syscall_handler.c
```diff
@@ -5,21 +5,15 @@ See the file LICENSE for details.
...
 
 #include "syscall.h"
-#include "console.h"
-#include "process.h"
+#include "sys_process_lifecycle.h"
+#include "kerneltypes.h"
+
+#include "console.h" // for debugging
 
-uint32_t sys_exit(uint32_t code) {
```
kkoser, line 11: Why did we get rid of these functions? They're kinda important

zachwaterson, line 11: They're moved into the process lifecycle file.

src/syscall.h
zachwaterson: Good catch; updated

src/sys_process_lifecycle.c
zachwaterson: Great catch; `kfree` is moved up to before the error checking.

src/sys_process_lifecycle.c
zachwaterson, line 82: Do we have an established error code system? I used -1 as a placeholder until then, since I didn't think it should be a blocking issue to getting this in.

src/sys_process_lifecycle.c
zachwaterson: Great thought; I do believe it's possible. Do we want to make this function atomic? Probably not. I'll update to pass the parent process in that inner function.

shuyangli:
Ryan says the interface for `run()` really wouldn't work for file system (see his comment below), so let's hold off on merging this. We'll discuss whether there's a better way to deal with permissions for file system.

rboccabe:
This scares me, or at least seems inefficient in that I'm passing my child's permissions to the run call explicitly. What this will require is for me to make a syscall which duplicates my own fs allowances linked list as a process, then I have to modify that linked list myself before passing it to the run call. Even if we create helpers to that modify the list, we've still given the parent access such that it can add a '/' allowance to its to-be-spawned child's allowances list. Now, run needs to ensure that each allowance is within the parent's allowances at the time run( ) is called. To me it makes more sense to ensure that the list is never in a bad state by using a pre-fork approach, where my to-be-spawned child's permissions are kept in my process struct and modified before I call run. These functions which modify the permissions do the security check and ensure the list is never in a bad state, provided that it starts in a good state (say, it starts with my parent's list of allowances to files, and I can further restrict that through a syscall interface before calling fork/exec or run). Imagine an interface with reset_future_child_permissions_to_mine( ), restrict_future_childs_permissions( ), and run_child( ). This will prevent failed run calls for illegal permissions, and inductively support our security by keeping everything in the kernel level rather than giving user level access and checking on run( ).

shuyangli:
So to implement fs restrictions with the current `run()` scheme, we need user-level `malloc`.

src/.gitignore
```diff
@@ -3,3 +3,4 @@ bootblock
 kernel
 basekernel.img
 basekernel.iso
+files.iso
```
shuyangli, line 6: Let's add a line to ignore `bin/*.nun`

src/global_structs.h
```diff
@@ -0,0 +1,39 @@
...
+#include "list.h"
+#include "kerneltypes.h"
+
+struct process_permissions {
+    struct list *directories;
+    int max_number_of_pages;
+    // todo: window stuff
+};
+
+struct process {
```
shuyangli, line 22: Why is `struct process` declaration in `global_structs.h`? None of the files that included `global_structs.h` (except `process.c`) uses `process`

src/syscall.h
```diff
@@ -8,24 +8,93 @@ See the file LICENSE for details.
...
+#define SYSCALL_debug_print 9000 // for debugging
 
+/**
+ * @brief   The main syscall function. All syscalls go through this.
+ * @details Use this function to implement the calling of the other syscalls
+ *          declared in syscall.h. Do not call it when the dedicated function
+ *          call for a specific syscall is available.
+ *
+ * @param   n The number of the syscall, declared in syscall.h
+ *
```
shuyangli, line 27: Can we get rid of the extra new lines in between `@param`s?

zachwaterson:
Closing for now, will reopen with the new permissions template code today

PR #56, opened 2016-04-07
File system syscalls

rboccabe:
This adds the syscalls for the filesystem. You are now able to open, write (and get rejected as we have no writable fs's implemented yet), read, close, and change permissions (only to levels of lesser privilege). Each process also has information about its access allowances in its process struct, and keeps track of its open files in a table. Error indicators have now been implemented as well.

This was tested using the same user level process as in the window/fs demo.

PS. Sorry about the commits. I tried an interactive rebase and apparently duplicated a bunch of commits. Any help there would be appreciated.
PPS. NUN-28 is outstanding on this PR


src/syscall.h
```diff
@@ -28,4 +35,27 @@ static inline int32_t yield() {
     return syscall(SYSCALL_yield, 0, 0, 0, 0, 0);
 }
```
rboccabe: Add documentation that points these functions to what eventually gets called.

src/fs_sys.c
```diff
@@ -0,0 +1,342 @@
...
+        return ERR_BAD_PATH;
+    }
+    if (path[2] != '/') {
+        return ERR_BAD_PATH;
+    }
+
+    char media_path[strlen(path) - 1];   //Two fewer than whole path length, + 1 for null terminator
+    strcpy(media_path, path+2);
+
+    int ata_unit = path[1] - 48; //48 is ascii for 0
```
kkoser: can we not just use '0' here to make it clearer?

src/fs_sys.c
```diff
@@ -0,0 +1,342 @@
...
+    }
+
+    struct fs_agnostic_file *fp = current->files->open_files[fd];
+    if (fp) {
+        switch (fp->ata_type) {
+            case 1:  //ISO
+                iso_fclose((struct iso_file *)fp);
+                break;
+            default:
+                break;
```
kkoser: Should we return an error indicating an invalid type here, just in case?

rboccabe: Good catch

src/fs_sys.c
rboccabe: Yup. Good idea.

src/fs_sys.c
```diff
@@ -0,0 +1,342 @@
...
+    //Set the lone permission in the list to root (that is all of the fs)
+    p->fs_allowances_head->next = 0;
+    strcpy(p->fs_allowances_head->path, "/");
+    p->fs_allowances_head->do_allow_below = 1;
+}
+
+// Look at the OS's table of open files to see if read write conflicts
+// might be occurring amongst many processes if you allow it to be opened
+bool fs_sys_check_open_conflict(const char *path, const char *mode) {
+    return 1;
```
kkoser: Should there be a todo here?

shuyangli:
I'm going to sit down with Ryan and clean up the commit history today. Please keep commenting, but don't merge before we rewrite history.

rboccabe:
That's sounds way more epic than it's going to be, @shuyangli

src/fs_sys.c
rboccabe: We can put one here, but I also made a JIRA issue [NUN-29] for having the OS keep an open files table. The problem is we can't really check it (or it will always return an okay) until we have a writable section of the fs which makes a conflict actually occur.

src/fs_sys_err.h
```diff
@@ -0,0 +1,23 @@
...
+See the file LICENSE for details.
+*/
+
+/* These will all be returned in syscalls which have a return type of unit32_t
+ * This means that they will all be very high constants because of two's
+ * complement on low absolute value negatives being nearest to the maximum
+ * value of 2^32. Hopefully this is far out of conflict with bytes_read
+*/
+
+#define ERR_FD_OOR -1   // file descriptor is out of range
```
kkoser: I thought we were going for positive error codes?

shuyangli: Negative; positive values are for valid return values.

rboccabe: Those would be very likely overlap with bytes read/written as the return value. These negatives being returned in uint32_t types will end up being very high positive values due to two's compliment, anyways.
What would you suggest as the error codes, instead?

shuyangli: Syscall return values should be `int32_t`, not `uint32_t`

src/syscall_handler.c
```diff
@@ -23,6 +24,31 @@ uint32_t sys_testcall(uint32_t code) {
     return 0;
 }
 
+uint32_t sys_open(uint32_t path_intptr, uint32_t mode_intptr) {
+    return fs_sys_open((const char *)path_intptr, (const char *)mode_intptr);
```
kkoser: How many levels of function wrapping do we want? Can't we just call the fs_sys\* in the switch for the handler, rather than putting another function on the stack just for casting?

src/fs_sys_err.h
kkoser: Fine with me, just checking. The window system calls follow the unix convention

rboccabe: So in syscall.h I return an int32_t for each of open/read/write/close, but the function syscall( ) itself returns a uint32_t, so anything below that might as well return a uint32_t as well.

shuyangli:
This is ready for merge; please keep reviewing.

src/Makefile
```diff
@@ -1,4 +1,4 @@
-OBJECTS=kernelcore.o main.o console.o $(MEMORY_OBJS) keyboard.o clock.o interrupt.o pic.o ata.o string.o graphics.o font.o syscall.o syscall_handler.o process.o mutex.o list.o pagetable.o rtc.o disk.o math.o window.o cmd_line.o $(TEST_OBJS) iso.o fs_terminal_commands.o
+OBJECTS=kernelcore.o main.o console.o $(MEMORY_OBJS) keyboard.o clock.o interrupt.o pic.o ata.o string.o graphics.o font.o fs_sys.o syscall.o syscall_handler.o process.o mutex.o list.o pagetable.o rtc.o disk.o math.o window.o cmd_line.o $(TEST_OBJS) iso.o fs_terminal_commands.o
```
shuyangli: Let's add another line of the form `OBJECTS+=blabla` to prevent merge conflicts for everyone.

rboccabe: Do we want to have a zillion OBJECTS+= lines just to prevent merge conflicts now and again, or can we just deal with them to avoid scrolling a long ways before we get to the substantial part of the makefile?

rboccabe: We'll take a vote with reactions to these comments, upvote the one you prefer.

PR #57, opened 2016-04-07
Add pointer mouse icon; fix mouse draw bug

alexmwu:
also fixes an bug that draws incorrectly on edges of the screen.

### DIFF
--- a/src/graphics.c
+++ b/src/graphics.c
@@ -74,7 +74,7 @@ static inline void plot_pixel(int x, int y, struct graphics_color c) {
 }
 
 // like plot_pixel, but does not do mouse erasing and redrawing or bounds checking
-static inline void draw_pixel(int x, int y, struct graphics_color c) {
+static inline void plot_pixel_no_check(int x, int y, struct graphics_color c) {
     if (x < 0 || x > video_xres - 1 || y < 0 || y > video_yres - 1) {
         return;
     }
@@ -139,7 +139,7 @@ void graphics_line(int x1, int y1, int x2, int y2, struct graphics_color c, bool
         int i;
         for (i = y1; i <= y2; i++) {
             if (mouse_draw) {
-                draw_pixel(x1, i, c);
+                plot_pixel_no_check(x1, i, c);
             }
             else {
                 plot_pixel(x1, i, c);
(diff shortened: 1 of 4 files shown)


src/graphics.c
```diff
@@ -68,10 +69,21 @@ static inline void plot_pixel(int x, int y, struct graphics_color c) {
         mouse_draw_buffer[index].r = c.r;
         mouse_draw_buffer[index].g = c.g;
         mouse_draw_buffer[index].b = c.b;
-        graphics_draw_mouse();
+        graphics_draw_pointer_mouse();
     }
 }
 
+// like plot_pixel, but does not do mouse erasing and redrawing or bounds checking
+static inline void draw_pixel(int x, int y, struct graphics_color c) {
```
jhamilt5: Might just be picky, but is there a better name we can use than `draw_pixel` since it seems fairly ambiguous with `plot_pixel` especially when used throughout the file.

src/graphics.c
```diff
@@ -237,28 +265,34 @@ void graphics_copy_to_color_buffer(int x, int y, int width, int height, struct g
...
-        v[1] = mouse_fg_color.g;
-        v[0] = mouse_fg_color.b;
-    }
+    graphics_line(mouse_x, start_y, mouse_x, end_y - 1, mouse_fg_color, 1);
+}
+
+// draws a standard pointer mouse
+void graphics_draw_pointer_mouse() {
+    // long pointer lines
+    // TODO: why does MOUSE_SIDE_2 - 3 work but not MOUSE_SIDE_2 - 1 not work (only draw line issue)
```
jhamilt5, line 283: Fix and remove the `TODO`

jhamilt5:
Functionality looks good to me. I like the mouse design. Just look at my line notes, but those were fairly low-priority in my opinion.

src/graphics.c
kkoser: I agree, this could get confusing. Lets call it something mouse to make it clear what its used for

src/graphics.c
alexmwu, line 283: I looked at this for a while and have no idea what's going on. I also talked with @kkoser about this, and we decided to just add a TODO.

src/graphics.c
alexmwu: `plot_pixel_direct`?

alexmwu: There could be other potential uses for this.

PR #58, opened 2016-04-07
[NUN-20] Fix Caps Lock Holds Down Shift

jhamilt5:
Caps lock only capitalizes characters. Caps lock does not modify any
other type of key. To modify other characters, you must hold down shift.


rboccabe:
Does the reverse work? That is when I have caps lock on and hold shift it goes back to lower case? This is the usual functionality but not necessary to include per se.

rboccabe:
Otherwise LGTM.

jhamilt5:
Yes, the reverse works.

alexmwu:
tested and it works!

PR #59, opened 2016-04-07
[NUN-31] Fix No 'W' Printing Bug

jhamilt5:
Fixes the inability to print the 'W' character. This was caused by the constant KEY_INVALID stomping on the ASCII value of 'W' which is 87. Fixed by assigning 'W' a temporary unique value and then handling that value later.
Tested by successfully printing all characters.

### DIFF
--- a/src/keyboard.c
+++ b/src/keyboard.c
@@ -12,9 +12,7 @@ See the file LICENSE for details.
 #include "kernelcore.h"
 #include "ps2.h"
 
-#define KEY_INVALID 0127
-
-#define W_SUBSTITUTE -1
+#define KEY_INVALID 0036
 
 #define SPECIAL_SHIFT 1
 #define SPECIAL_ALT   2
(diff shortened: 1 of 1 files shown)


src/keyboard.c
```diff
@@ -14,6 +14,8 @@ See the file LICENSE for details.
 
 #define KEY_INVALID 0127
 
+#define W_SUBSTITUTE -1
```
shuyangli: This looks super sketchy... I'm trying to figure out if we can substitute the value of `KEY_INVALID` with `-1` or maybe `255`, do you have any insight?

jhamilt5: When we switch the key invalid value, it will just stomp on another valid ascii value. I experimented with several values, but the mapping was odd, and it's hard to find the correlation of key mapping to ascii values. 
I agree it's sketchy, but I also think it's akin to putting plywood over that sketchy hole of memory

shuyangli: I see that, and that's why I proposed 255 (W is apparently a type-able char, can we type ASCII 255 from keyboard?) If there's no better alternative, I'm happy to merge this.

src/keyboard.c
```diff
@@ -95,12 +97,32 @@ static char keyboard_map(int code) {
             return KEY_INVALID;
         } else if (shiftlock_mode) {
             if (shift_mode) {
-                return keymap[code].normal;
+                if (keymap[code].normal >= 97 && keymap[code].normal <= 122) {
+                    return keymap[code].normal;
+                }
+                else {
```
shuyangli: Style: all `else`'s should be on the same line as the closing curly brace of the `if` block

rboccabe:
@jhamilt5 After a fixing merge conflicts and Shuyang's style fix you should be good to merge yourself, just be sure to make and run, then wait for Travis CI.

kkoser:
tested and confirmed. LGTM!

PR #61, opened 2016-04-12
Add permission template and process execution syscalls

zachwaterson:
We're back with an even more monstrous commit. The module for permissions templates is created, and the process lifecycle syscalls are updated to use it. This commit also reflects the new code organization schema for syscalls and creates a binary folder for us to store user-level binaries. These binaries will be automatically copied over into files.iso. 

Known todos: the run() syscall requires that the executable be in ATA 3; this will need to be made more robust by people who know the filesystem better than me. 

We also need to make it so killing a process kills its children, but I want to talk to Thain about the best way to do that first.


src/permissions_template.c
```diff
@@ -0,0 +1,119 @@
...
+#include "permissions_template.h"
+#include "console.h"
+
+#define MAX_NUMBER_OF_PERMISSIONS 169 // temporary, for now they'll be statically allocated
+
+static uint32_t template_counter = 1;
+static struct permissions_template* permissions_table[MAX_NUMBER_OF_PERMISSIONS]; // statically allocated array of templates
+
+uint32_t generate_new_identifier() {
+    return template_counter++;
```
kkoser: We should discuss a way to hash these identifiers or similar. If they follow a predictable pattern, it opens the possibility for processes to request modifications to permissions that they did not create.

src/permissions_template.c
```diff
@@ -0,0 +1,119 @@
...
+}
+
+struct process_permissions* permissions_from_template(struct permissions_template *template) {
+
+    if (template <= 0) {
```
kkoser: Why < 0? Its a pointer.

src/permissions_template.c
```diff
@@ -0,0 +1,119 @@
...
+        console_printf("Error: invalid template identifier\n");
```
kkoser: Not really a correct error message here? Its the template that invalid, not necessarily the identifier

src/permissions_template.c
```diff
@@ -0,0 +1,119 @@
...
+    // log it in the table
+    uint32_t identifier = generate_new_identifier();
+    permissions_table[identifier] = new_template;
+
+    // set default values to be related to parents
+    new_template->directories = current->permissions->directories;
+    new_template->max_number_of_pages = current->permissions->max_number_of_pages / 3;
+    new_template->max_width = current->permissions->max_width;
+    new_template->max_height = current->permissions->max_height;
+    new_template->offset_x = current->permissions->offset_x;
```
kkoser: For windowing, all offsets are relative to the parent. We should default to an offset of 0 in permissions

src/permissions_template.h
```diff
@@ -0,0 +1,97 @@
...
+    // Window permissions, to override parents'
+    int max_width;
+    int max_height;
+    int offset_x;
+    int offset_y;
+};
+
+/**
+ * @brief   Creates a new permissions template for a process to use.
+ * @details Allocates an identifier for a process permissions template,
```
kkoser: np: phrasing. We aren't allocating an identifier, but rather a permissions template

src/permissions_template.h
```diff
@@ -0,0 +1,97 @@
...
+ *          memory allocation.
+ *
+ * @return  The identifier of the newly created process identifier.
+ */
+uint32_t create_permissions_template();
+
+/**
+ * @brief   Deletes a process permissions template.
+ * @details Deletes the process permissions template associated with the identifier.
+ *          This function does not check the ownership of the template before deleting.
```
kkoser: We should put this in big letters or something to make it clear

src/process.c
```diff
@@ -33,6 +35,17 @@ void process_init() {
 
     current->state = PROCESS_STATE_READY;
 
+    // establish initial permissions
+    struct process_permissions *initial_permissions = kmalloc(sizeof(struct process_permissions));
+    initial_permissions->max_number_of_pages = memory_pages_total();
+    // initial_permissions.directories =
+    initial_permissions->max_width = 640; // temporary
```
kkoser: We should use video_xres()/yres() here

src/process.h
```diff
@@ -12,13 +12,28 @@ See the file LICENSE for details.
 #include "pagetable.h"
 #include "x86.h"
 #include "memory.h"
+#include "syscall.h"
```
kkoser: Why does this need to include sys call.h?

src/syscall_handler.c
```diff
@@ -28,10 +23,17 @@ int32_t syscall_handler(uint32_t n, uint32_t a, uint32_t b, uint32_t c,
     switch (n) {
         case SYSCALL_exit:
             return sys_exit(a);
-        case SYSCALL_testcall:
-            return sys_testcall(a);
         case SYSCALL_yield:
             return sys_yield();
+        case SYSCALL_run:
+            return sys_run((char *)a, b, current);
```
kkoser, line 32: const char *? Goes for function decl as well

kkoser:
We need to decide what our system calls are returning. We currently have the handle functions returning uint32, but the functions in sys call.h (user level) return signed integers. If we want to be returning negative values to indicate errors, we should probably switch to signed integers across the board.

src/permissions_template.c
zachwaterson: Currently, there is a permissions check function can be used to prevent access to non-owned templates. I looked into using a dictionary or something but the complexity didn't seem worth it. Do you think we should still do more?

src/permissions_template.c
zachwaterson: Good catch; this was leftover from the pre-pointer implementation

src/permissions_template.c
zachwaterson: Good point. Fixed!

src/permissions_template.c
zachwaterson: Done!

src/permissions_template.h
zachwaterson: Correct; good catch

src/permissions_template.h
zachwaterson: The lack of deletion checking? alrighty

src/process.c
zachwaterson: They weren't public and I wanted a second opinion on whether to make them such; now I shall

zachwaterson: For some reason I thought they weren't public; thanks for pointing that out

src/process.h
zachwaterson: This was @shuyangli 's doing, he can explain

src/syscall_handler.c
zachwaterson, line 32: sure

zachwaterson:
Agreed; I want to leave that to another pull request, though, since this one is 600 lines already...

src/permissions_template.c
kkoser: I think it would be useful, but something that we can tackle at a later date. We will eventually need hashing functions for other things (passwords and hash tables for instance) so when we add that we can update this. Just something to think about

kkoser:
This is an issue I think we need to resolve now since this commit already touched all existing syscalls and will set an example for all the ones going forward

src/process.c
alexmwu: they aren't. it's graphics_width() and height() I believe

zachwaterson: Yep; updated to use the functions

zachwaterson:
Updated syscalls to return signed ints, let's look at it now

alexmwu:
@zachwaterson can you add separate JIRA issues to dynamically allocate the templates array and hash identifiers?

src/permissions_template.c
```diff
@@ -0,0 +1,119 @@
...
+    new_permissions->offset_x = template->offset_x;
+    new_permissions->offset_y = template->offset_y;
+
+    // todo: the rest of the permissions
+
+    return new_permissions;
+}
+
+struct permissions_template* template_for_identifier(uint32_t identifier) {
+    if (identifier < 0) {
```
alexmwu: identifier is unsigned

zachwaterson: Good catch

src/bin/test_userproc_even.c
```diff
@@ -0,0 +1,29 @@
+int main();
+int _start() {
+    return main();
+
+    // invoke system call that kills the process
+}
+
+#include "../syscall.h"
```
kkoser: Should we be moving the executable files one directory up or something? This seems weird

zachwaterson: The Makefile is built around the executables being in bin/; we'd have to rework that if we want to get rid of the relative path. The Makefile also automatically moves compiled files in bin/ to the files.iso. I kind of like having them in their own folder to emphasize that they have different compiler rules and that they get copied to the ISO

src/permissions_template.c
```diff
@@ -0,0 +1,115 @@
...
+    }
+    // set its owner
+    new_template->owner = current;
+    // log it in the table
+    uint32_t identifier = generate_new_identifier();
+    permissions_table[identifier] = new_template;
+
+    // set default values to be related to parents
+    new_template->directories = current->permissions->directories;
+    new_template->max_number_of_pages = current->permissions->max_number_of_pages / 3;
```
alexmwu: the magic number 3 should probably be documented somewhere (Doxygen) or set to some global constant

src/permissions_template.c
```diff
@@ -0,0 +1,115 @@
...
+
+
+#include "kerneltypes.h"
+#include "kmalloc.h"
+#include "permissions_template.h"
+#include "console.h"
+
+#define MAX_NUMBER_OF_PERMISSIONS 169 // temporary, for now they'll be statically allocated
+
+static uint32_t template_counter = 1;
```
alexmwu: really np: any reason we start at 1 and not 0?

src/permissions_template.c
```diff
@@ -0,0 +1,115 @@
...
+    new_template->offset_x = 0;
+    new_template->offset_y = 0;
+    return identifier;
+
+}
+
+void delete_templates_owned_by_process(struct process *p) {
+    int i;
+    for (i = 1; i < template_counter; i++) {
+        struct permissions_template *t = template_for_identifier(i);
```
alexmwu: null check here and set the pointer in the table to null (in delete_permissions_template). otherwise we could deref 0 or a dangling pointer later on.

src/permissions_template.c
zachwaterson: I didn't want the user to be confused if they get 0 back, since 0 is often associated with failure. It also helps me keep my code straight, since I check for 0 on return often.

src/permissions_template.c
zachwaterson: Great catch; updated

src/bin/test_userproc_even.c
```diff
@@ -0,0 +1,29 @@
...
+
+int main() {
+    int has_spawned = 0;
+    while(1) {
```
rboccabe: np: space after while

src/bin/test_userproc_even.c
```diff
@@ -0,0 +1,29 @@
...
+        int i = 0;
+        for(i = 0; i < 10; i+=2) {
```
rboccabe: np: space after for

src/bin/test_userproc_even.c
```diff
@@ -0,0 +1,29 @@
...
+            debug_print(i);
+            int j;
+            for(j = 0; j < 100000000; j++) {} // busywait
```
rboccabe: np: space after for

src/memory_raw.c
```diff
@@ -13,15 +13,15 @@ See the file LICENSE for details.
...
-
 static uint32_t *freemap = 0;
 static uint32_t freemap_bits = 0;
 static uint32_t freemap_bytes = 0;
 static uint32_t freemap_cells = 0;
 static uint32_t freemap_pages = 0;
 
+static uint32_t pages_free = 0;
+static uint32_t pages_total = 0;
+
```
rboccabe: Why are we moving these? It shouldn't change anything butit doesn't look like we need this file being changed on this PR.

src/pagetable.c
```diff
@@ -252,10 +252,20 @@ void exception_handle_pagefault(int intr, int code) {
             interrupt_dump_process();
         }
 
-        // Currently we give it as much memory as we could
-        pagetable_alloc(current->pagetable, vaddr, PAGE_SIZE,
+        int number_of_pages_left = current->permissions->max_number_of_pages - current->number_of_pages_using;
+        // check the permissions of the process
+        if(number_of_pages_left > 0) {
```
rboccabe: np: space after if

src/permissions_template.c
```diff
@@ -0,0 +1,117 @@
...
+
+#include "kerneltypes.h"
+#include "kmalloc.h"
+#include "permissions_template.h"
+#include "console.h"
+
+#define MAX_NUMBER_OF_PERMISSIONS 169 // temporary, for now they'll be statically allocated
+#define DEFAULT_FRACTION_OF_MEMORY 3
+
+static uint32_t template_counter = 1;
```
rboccabe: Will we be missing index 0 if we start this at 1?
Furthermore, if we are always calling `generate_new_identifier` will our first identifier end up being 2?

zachwaterson: The postfix returns the original value, so the first identifier will be 1. This is intentional; see my response to Alex's comment

src/permissions_template.c
```diff
@@ -0,0 +1,117 @@
...
+        return new_permissions;
+    }
+    // apply information
+    new_permissions->max_number_of_pages = template->max_number_of_pages;
+    new_permissions->max_width = template->max_width;
+    new_permissions->max_height = template->max_height;
+    new_permissions->offset_x = template->offset_x;
+    new_permissions->offset_y = template->offset_y;
+
+    // todo: the rest of the permissions
```
rboccabe: We'll have to be careful with the fs permissions here in the future, as it looks like we'll need to copy a linked list as opposed to just copying a pointer (the parent could then delete the template!).

src/permissions_template.c
```diff
@@ -0,0 +1,117 @@
...
+
+struct permissions_template* template_for_identifier(uint32_t identifier) {
+    if (identifier >= template_counter) {
+        console_printf("Error: invalid identifier\n");
+        return 0; // error case, todo
+    }
+    return permissions_table[identifier];
+}
+
+struct process_permissions* permissions_from_identifier(uint32_t identifier) {
```
rboccabe: This is a beautiful function. +1

src/permissions_template.c
```diff
@@ -0,0 +1,117 @@
...
+
+    // set default values to be related to parents
+    new_template->directories = current->permissions->directories;
+    new_template->max_number_of_pages = current->permissions->max_number_of_pages / DEFAULT_FRACTION_OF_MEMORY;
+    new_template->max_width = current->permissions->max_width;
+    new_template->max_height = current->permissions->max_height;
+    new_template->offset_x = 0;
+    new_template->offset_y = 0;
+    return identifier;
+
```
rboccabe: np: extra line

src/permissions_template.c
```diff
@@ -0,0 +1,117 @@
...
+    uint32_t identifier = generate_new_identifier();
+    permissions_table[identifier] = new_template;
+
+    // set default values to be related to parents
+    new_template->directories = current->permissions->directories;
+    new_template->max_number_of_pages = current->permissions->max_number_of_pages / DEFAULT_FRACTION_OF_MEMORY;
+    new_template->max_width = current->permissions->max_width;
+    new_template->max_height = current->permissions->max_height;
+    new_template->offset_x = 0;
+    new_template->offset_y = 0;
```
rboccabe: Add a todo here for fs stuff?

src/permissions_template.c
```diff
@@ -0,0 +1,117 @@
...
+    }
+}
+
+void delete_permissions_template(uint32_t identifier) {
+    struct permissions_template *template = template_for_identifier(identifier);
+    permissions_table[identifier] = 0; // clear out of the table
+    kfree(template);
+}
+
+// void initialize_permissions_table() {
```
rboccabe: Should we eliminate the commented out code? Or does it serve a purpose?

src/permissions_template.h
```diff
@@ -0,0 +1,97 @@
...
+ * @brief   Creates a set of process_permissions corresponding to a permissions template.
+ * @details Using the information stored in the template, creates and returns a pointer to a
+ *          process_permissions struct to be associated with a process. It will need to be
+ *          freed upon conclusion of use (typically when the process storing the reference to it ends).
+ *
+ * @param   identifier The identifier for the template to be used for permissions
+ *          creation.
+ * @return  The process_permissions struct with the necessary information.
+ */
+struct process_permissions* permissions_from_identifier(uint32_t identifier);
```
rboccabe: np: asterisk goes with function name, not struct name

src/permissions_template.h
```diff
@@ -0,0 +1,97 @@
...
+ * @details Looks up and returns a pointer to the template in the table. This
+ *          function is for kernel use only, for other modules to use to
+ *          manipulate the template struct. Users will manipulate it through
+ *          other syscalls. This function does not check the ownership of the
+ *          template before returning.
+ *
+ * @param   identifier The identifier for the template to return.
+ * @return  A pointer to the desired template, or 0 if not found.
+ */
+struct permissions_template* template_for_identifier(uint32_t identifier);
```
rboccabe: np: asterisk goes with function name, not struct name

src/permissions_template.c
```diff
@@ -0,0 +1,117 @@
...
+    new_permissions->max_height = template->max_height;
+    new_permissions->offset_x = template->offset_x;
+    new_permissions->offset_y = template->offset_y;
+
+    // todo: the rest of the permissions
+
+    return new_permissions;
+}
+
+struct permissions_template* template_for_identifier(uint32_t identifier) {
```
rboccabe: np: asterisk goes with function name, not struct name

src/permissions_template.c
rboccabe: np: asterisk goes with function name, not struct name

src/permissions_template.h
```diff
@@ -0,0 +1,97 @@
...
+ *          owned by the given process.
+ *
+ * @param   identifier The identifier of the template to check.
+ * @param   process The process to check ownership of.
+ *
+ * @return  1 if the process owns the template and 0 otherwise.
+ */
+int template_owned_by_process(uint32_t identifier, struct process *p);
+
+// void initialize_permissions_table();
```
rboccabe: Same comment as above on commented out code.

src/process.c
```diff
@@ -33,6 +36,17 @@ void process_init() {
...
     current->state = PROCESS_STATE_READY;
 
+    // establish initial permissions
+    struct process_permissions *initial_permissions = kmalloc(sizeof(struct process_permissions));
+    initial_permissions->max_number_of_pages = memory_pages_total();
+    // initial_permissions.directories =
+    initial_permissions->max_width = graphics_width();
+    initial_permissions->max_height = graphics_height();
+    initial_permissions->offset_x = 0;
+    initial_permissions->offset_y = 0;
```
rboccabe, line 46: Add an fs todo?

src/sys_permissions.h
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * @brief   Creates a new permissions template for a process to use.
+ * @details Allocates an identifier for a process permissions template,
+ *          to be adjusted as needed before passed as a parameter to the
+ *          run() syscall. The identifier will need to be freed by the process
+ *          when no longer needed so as to stop counting toward the process'
+ *          memory allocation.
+ *
+ * @return  The identifier of the newly created process identifier.
```
rboccabe: This sounds like a pid. "of the newly created permissions template"?

src/sys_process.h
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * @brief   Creates and begins execution of a new process.
+ * @details Loads a new process into memory as a child of the current process,
+ *          and then begins execution of that process. The current process
+ *          yields and is placed at the end of the ready queue.
+ *
+ * @param   process_path The filesystem path of the process executable, which should be
+ *          a .nun executable.
+ *
+ * @param   child_permissions The set of permissions to be passed to the child,
```
rboccabe: permissions_identifier

src/syscall_handler_permissions.c
```diff
@@ -0,0 +1,19 @@
+/*
+Copyright (C) 2015 The University of Notre Dame
+This software is distributed under the GNU General Public License.
+See the file LICENSE for details.
+*/
+
+#include "kerneltypes.h"
+#include "permissions_template.h"
```
rboccabe: It compiles but you're not including syscall_handler_permission.h.

src/permissions_template.c
zachwaterson: This is the scaffolding for a dynamically allocated permissions table; I can get rid of it for now

src/sys_permissions.h
zachwaterson: Yep, good catch

src/syscall_handler_permissions.c
zachwaterson: I'll include it for good measure

src/bin/test_userproc_even.c
kkoser: Would it make some sense to have the makefile copy syscall.h into that directory? Or just have it live there entirely and have the nunya.iso make reference it one directory up? I think it would be more annoying to have the ../ in every user program we write than one weird line in the makefile

dthain: The customary way to solve this is to set up a directory that contains the "public" includes, then tell the compiler to make it part of the include path with the `-I` option.  That way, user-level code doesn't need to know anything about the filesystem layout, it just accepts what the compiling environment has set up.

alexmwu:
did a run over and LGTM

PR #62, opened 2016-04-15
Open Files Table

jhamilt5:
Restructure the process struct and fs.c layout to create a universal list of opened files for the kernel to view. Allocates 1024 fs_agnostic files to the kernel and initializes each of them to closed. Every process has its own set of file descriptors which reference some member of the kernel's list of opened/unopened files. 
Security is unchanged from Ryan's fs_syscall branch.
In fact, this PR includes all of Ryan's code for the fs_syscall PR, so merging this would integrate his branch as well.


src/fs.c
```diff
@@ -0,0 +1,303 @@
...
+void fs_init_security(struct process *p) {
+    // Initialize all file descriptor slots to 0
+    int i;
+    for (i = 0; i < PROCESS_MAX_OPEN_FILES; i++) {
+        p->fd_table[i].is_open = 0;
+        p->fd_table[i].ptr = 0;
+    }
+
+    //Allowances alloc'ing.
+    struct fs_allowance *allowance = kmalloc(sizeof(*allowance));
```
shuyangli, line 53: Did we ever free the allowance?

src/fs.c
```diff
@@ -0,0 +1,303 @@
...
+    allowance->do_allow_below = 1;
+
+    //Set the lone permission in the list to root (that is all of the fs)
+    list_push_head(&(p->fs_allowances_list), (struct list_node *)allowance);
+}
+
+// Look at the OS's table of open files to see if read write conflicts
+// might be occurring amongst many processes if you allow it to be opened
+bool fs_check_open_conflict(const char *path, const char *mode) {
+    return 1;
```
shuyangli, line 64: Please add a JIRA issue for this

src/fs.c
```diff
@@ -0,0 +1,303 @@
...
+}
+
+int map_media_to_driver_id(int media) {
+    return ISO;
+}
+
+int mode_str_to_int(const char *mode) {
+    if (strcmp(mode, "a") == 0) {
```
shuyangli, line 72: Can we be a little smarter than this? Something like `if "a" in mode then flag |= APPEND`

src/fs.c
```diff
@@ -0,0 +1,303 @@
...
+    }
+
+    //Allowances alloc'ing.
+    struct fs_allowance *allowance = kmalloc(sizeof(*allowance));
+    strcpy(allowance->path, "/");
+    allowance->do_allow_below = 1;
+
+    //Set the lone permission in the list to root (that is all of the fs)
+    list_push_head(&(p->fs_allowances_list), (struct list_node *)allowance);
+}
```
jhamilt5, line 59: I think we should wrap this into a nicely named function like `add_allowance(struct fs_allowance *allowance)`
Similarly, we should add a function `remove_allowance()` which is where we would `kfree()` the allowance, @shuyangli

src/fs.c
jhamilt5, line 72: If we had a `contains(char c)` function in `string.h` I would say yes, but this looks fine to me, especially because this is every possible case already enumerated.

src/fs.c
rboccabe, line 59: The syscalls which add or remove allowances will modify a permissions template through a permissions template identifier, rather than doing it to the current process (or the process being passed in). That is, once you have your set of file allowances as a process, it should not change.

src/fs.c
```diff
@@ -0,0 +1,303 @@
...
+    return 0;
+}
+
+struct fs_agnostic_file *create_fs_agnostic_file(enum ata_kind ata_type, uint8_t ata_unit, const char *path, uint8_t mode) {
+    int next_open_slot;
+    for (next_open_slot = 0; next_open_slot < MAX_OS_OPEN_FILES; next_open_slot++) {
+        if (!open_files_table[next_open_slot].filep) {  // If the file pointer hasn't been assigned, this is an unused file slot
+            break;
+        }
+    }
```
jhamilt5, line 126: Need to handle the case where there are no open file slots

src/fs.c
rboccabe, line 72: It ignores the technically legal but to me always confusing "wa", "aw", "raw", etc. I can add those if we want to support them.

shuyangli, line 72: ^ because of this I think we should make it right (add a `contains` function in `string.h`), because we can't guarantee the users won't write things like `wa` or `aw`

src/fs.c
```diff
@@ -0,0 +1,303 @@
...
+        if (path_permitted_by_allowance(path, (struct fs_allowance *)iterator)) {
+            return 1;
+        } else {
+            iterator = iterator->next;
+        }
+    }
+    return 0;
+}
+
+uint32_t fs_security_check(const char *path) {
```
shuyangli: Shouldn't this be signed?

src/fs.c
```diff
@@ -0,0 +1,303 @@
...
+        kfree(new_file);
+        return 0;
+    }
+    return new_file;
+}
+
+int32_t fs_open(const char *path, const char *mode) {
+    //checks both owner permissions (dummy true) and process permissions
+    uint32_t success = fs_security_check(path);
+    if (success < 1) {
```
shuyangli, line 151: As `fs_security_check` should return a signed value, not sure if C would convert return values automatically, but for correctness sake `success` should be signed.

src/fs.c
rboccabe: Indeed.

src/fs.c
rboccabe, line 151: Indeed.

src/fs.c
```diff
@@ -0,0 +1,303 @@
...
+           ) {
+            // The allowance is above the requested path
+            return 1;
+        } else {
+            return 0;
+        }
+    } else {
+        //in this case, you're allowed to read dir contents
+        //but not access them
+        if (strcmp(allowed->path, path) == 0) {
```
shuyangli, line 294: `return (strcmp(allowed->path, path) == 0);`

rboccabe, line 294: This is fine as is as they are functionally equivalent.

shuyangli:
This looks good for the most part, ideally we should also build `remove_all_allowance()` so we can clean up properly on process exit.

PR #63, opened 2016-04-18
Add clock and rtc syscalls

alexmwu:
There is a bug with the clock as well (described in [NUN-39]).


src/bin/test_clock.c
```diff
@@ -0,0 +1,59 @@
+int main();
+int _start() {
+    return main();
+
+    // invoke system call that kills the process
+}
+
+#include "../syscall.h"
```
shuyangli: this shouldn't need the "../" part with the current makefile

src/bin/test_clock.c
```diff
@@ -0,0 +1,59 @@
+int main();
+int _start() {
+    return main();
+
+    // invoke system call that kills the process
+}
+
+#include "syscall.h"
+#include "../clock.h"
```
shuyangli: `clock_struct.h?` and we should move all the user-accessible structs into `/include`

src/Makefile
```diff
@@ -12,8 +13,11 @@ DEBUG_OBJS = debug_kernel.o
 FS_OBJS = fs.o syscall_handler_fs.o
 WINDOW_OBJS = window.o graphics.o syscall_handler_window.o window_manager.o
 PROCESS_OBJS = syscall_handler_process.o syscall_handler_permissions.o process.o permissions_capabilities.o
+CLOCK_OBJS = syscall_handler_clock.o syscall_handler_rtc.o
+
+BINARIES = bin/test_userproc_even.nun bin/test_window.nun bin/test_clock.nun
```
kkoser, line 18: Why are we defining binaries twice? Looks like a missed line from a merge?

kkoser:
Clock test does not work, fails accessing binary. This is likely due to the redefinition of BINARIES in the makefile

alexmwu:
the test worked for me without fixing the redefinition, but hopefully it works in general now.

kkoser:
Looks good now, ship it!

PR #64, opened 2016-04-22
Window events

kkoser:
This adds a preliminary implementation of the event system for windows.
Events will be captured and added to an event queue for the active
window. The process can retrieve these events in the order that they
came in using the get_event system call. They provide a buffer into
which the event information should be copied.

If there is no active window, events will be distributed directly to
the kernel console.

### DIFF
--- a/src/keyboard.c
+++ b/src/keyboard.c
@@ -11,6 +11,7 @@ See the file LICENSE for details.
 #include "process.h"
 #include "kernelcore.h"
 #include "ps2.h"
+#include "window_manager.h"
 
 #define KEY_INVALID 0036
 
@@ -130,9 +131,16 @@ void keyboard_interrupt(int i, int code) {
     if ((buffer_write + 1) == (buffer_read % KEYBOARD_BUFFER_SIZE)) {
         return;
     }
-    buffer[buffer_write] = c;
-    buffer_write = (buffer_write + 1) % KEYBOARD_BUFFER_SIZE;
-    process_wakeup(&queue);
+    if (active_window) {
+        send_event_keyboard_press(c);
+    } else {
+        if ((buffer_write + 1) == (buffer_read % KEYBOARD_BUFFER_SIZE)) {
+            return;
+        }
+        buffer[buffer_write] = c;
+        buffer_write = (buffer_write + 1) % KEYBOARD_BUFFER_SIZE;
+        process_wakeup(&queue);
+    }
 }
 
 char keyboard_read() {
--- a/src/keyboard.c
+++ b/src/keyboard.c
@@ -128,9 +128,6 @@ void keyboard_interrupt(int i, int code) {
         return;
     }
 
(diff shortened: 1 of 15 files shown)


src/keyboard.c
```diff
@@ -130,9 +131,16 @@ void keyboard_interrupt(int i, int code) {
     if ((buffer_write + 1) == (buffer_read % KEYBOARD_BUFFER_SIZE)) {
         return;
     }
-    buffer[buffer_write] = c;
-    buffer_write = (buffer_write + 1) % KEYBOARD_BUFFER_SIZE;
-    process_wakeup(&queue);
+    if (active_window) {
+        send_event_keyboard_press(c);
+    } else {
+        if ((buffer_write + 1) == (buffer_read % KEYBOARD_BUFFER_SIZE)) {
```
rboccabe, line 134: duplicated from line 130 outside of the if

src/sys_window.h
```diff
@@ -124,5 +124,8 @@ static inline int32_t draw_string(int x, int y, const char *str, const struct gr
 	return syscall(SYSCALL_window_draw_string, x, y, (int)str, (int)fgcolor, (int)bgcolor);
 }
 
+static inline int32_t get_event(struct event *e) {
+	return syscall(SYSCALL_window_get_event, (int)e, 0, 0, 0, 0);
```
rboccabe: uint32_t for argument e?

src/syscall_handler_window.c
```diff
@@ -52,4 +53,17 @@ int32_t sys_draw_string(int x, int y, const char *str, const struct graphics_col
...
-}
\ No newline at end of file
+}
+
+int32_t sys_get_event(struct event *e) {
+    CHECK_PROC_WINDOW();
+    struct list *list = &(current->window->event_queue);
+    struct event *last_event = (struct event *)list_pop_tail(list);
+    if (last_event == 0) {
+        return 2;
```
rboccabe, line 63: What's this magical 2 mean?

src/window_manager.h
```diff
@@ -0,0 +1,20 @@
...
+This software is distributed under the GNU General Public License.
+See the file LICENSE for details.
+*/
+
+#ifndef WINDOW_MANAGER_H
+#define WINDOW_MANAGER_H
+
+#include "sys_window_struct.h"
+
+void send_event_mouse_click();
```
rboccabe, line 17: Lacking Doxygen comments.

rboccabe:
Outside of those few comments LGTM.

src/syscall_handler_window.c
kkoser, line 63: This is to indicate that there is no event. Added to documentation

src/window_manager.h
kkoser, line 17: Added

src/sys_window.h
kkoser: Sure

PR #66, opened 2016-04-29
Fix potential window concurrency issues

alexmwu:
remove static bounds variables in graphics.h and static offset variables in window.h
moved to struct window


src/window.c
```diff
@@ -11,19 +11,33 @@ See the file LICENSE for details.
...
+}
 
 // This function prepares to draw in a window by setting up its boundary
 // restrictions.
-// THIS FUNCTION SHOULD ALWAYS BE CALLED BEFORE MAKING A GRAPHICS DRAW
-// CALL IN THE WINDOW SYSTEM
-static inline void window_begin_draw(struct window *w) {
+// THIS FUNCTION SHOULD ALWAYS BE CALLED WHEN MAKING A WINDOW
+// OR MOVING A WINDOW
+static inline void window_init_draw(struct window *w) {
```
kkoser, line 34: why the function name change here?

src/window.c
```diff
@@ -63,13 +66,19 @@ struct window *window_create(int x, int y, int width, int height, struct window
     w->parent = parent;
     struct list l = {0, 0};
     w->event_queue = l;
+    w->x_offset = 0;
+    w->y_offset = 0;
+    w->bounds_x_1 = 0;
+    w->bounds_x_2 = 0;
+    w->bounds_y_1 = 0;
+    w->bounds_y_2 = 0;
+	window_init_draw(w);
```
kkoser: Please use spaces

src/window.c
```diff
@@ -133,22 +132,3 @@ void window_draw_string(struct window *w, int x, int y, const char *str, struct
     }
 }
 
-void window_draw_bitmap(struct window *w, int x, int y, int width, int height, uint8_t * data,
```
kkoser, line 138: Why did you just remove this functionality?

src/window.c
alexmwu, line 34: begin to me implies that there should be an end. init does not say the same thing. It just says that we initialize the window's ability to draw.

src/window.c
alexmwu, line 138: good catch. my b

kkoser:
Builds and things appear to run.

zachwaterson:
Just looked over it, and LGTM!

PR #67, opened 2016-04-29
Added window clear

kkoser:
This adds a window clearing to system calls, enabling people to create cleaner
looking programs. This is accomplish using the graphics_rect function to set
the window area to the window's background color.

Tested by adding functionality to clear after a certain number of characters
have been entered in the window_test function

### DIFF
--- a/src/bin/test_window.c
+++ b/src/bin/test_window.c
@@ -13,12 +13,19 @@ int main() {
     struct graphics_color bg = {0,0,0};
     struct event e;
     int x = 0;
+    int count = 0;
     while (1) {
         int res = get_event(&e);
         if (res == 0) {
+            if (count > 10) {
+                clear();
+                count = 0;
+                x = 0;
+            }
             if (e.type == EVENT_KEYBOARD_PRESS) {
                 draw_char(x, 0, e.character, &c, &bg);
                 x += 10;
+                count++;
             }
         }
     }
--- a/src/window.c
+++ b/src/window.c
@@ -66,6 +66,8 @@ struct window *window_create(int x, int y, int width, int height, struct window 
 
     // Draw the border
     struct graphics_color bc = {128,128,128};
+    struct graphics_color background = {0,0,0};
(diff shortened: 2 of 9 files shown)


src/window.c
```diff
@@ -66,6 +66,8 @@ struct window *window_create(int x, int y, int width, int height, struct window
 
     // Draw the border
     struct graphics_color bc = {128,128,128};
```
shuyangli: np: Is this supposed to be "border color?" (can we choose a descriptive name, since the acronym for background color is also bc)

src/bin/test_window.c
```diff
@@ -13,12 +13,19 @@ int main() {
...
     while (1) {
         int res = get_event(&e);
         if (res == 0) {
+            if (count > 10) {
+                clear();
+                count = 0;
+                x = 0;
+            }
             if (e.type == EVENT_KEYBOARD_PRESS) {
                 draw_char(x, 0, e.character, &c, &bg);
```
shuyangli, line 26: Right now we can draw directly onto the border, and when we clear we won't clear the bits that were drawn. Do we want the border to be part of the window? This is relatively minor, we don't need to fix it now.

screen shot 2016-05-03 at 16 41 40

shuyangli:
looks good other than the above minor issue

src/bin/test_window.c
kkoser, line 26: Correct, this is an issue of whether the window border is a part of the window (in which case it makes sense to be able to draw on top of them) or outside the window (in which case, does the width of border count against the size of the window?). I would rather get the clear function in right now and tackle these larger issues later.

src/window.c

src/bin/test_window.c
alexmwu, line 26: Whichever way we do it, we shouldn't let the user modify the border other than the color

shuyangli, line 26: I think what we have suits for now

PR #68, opened 2016-05-03
Moved math.h into user include

kkoser:


zachwaterson:
Makes sense to me! and it doesn't break any existing paths

PR #69, opened 2016-05-03
Memory syscalls

zachwaterson:
This does a few things:
1) Adds the syscalls for altering a capability's memory restrictions
2) creates a demo for those restrictions (memory_demo)
3) fixes the zombie process bug
4) Add code to clean up processes in general


src/bin/print_even.c
```diff
@@ -0,0 +1,19 @@
...
+
+int main() {
+    int i = 0;
+    for (i = 0; i < 11; i+=2) {
+        debug_print(i);
+        int j;
+        for (j = 0; j < 500000000; j++) {} // busywait
+    }
+
+    exit(0);
```
kkoser, line 16: Why do we exit and return?

src/bin/print_odd.c
```diff
@@ -0,0 +1,19 @@
...
+
+int main() {
+    int i = 0;
+    for (i = 1; i < 11; i+=2) {
+        debug_print(i);
+        int j;
+        for (j = 0; j < 500000000; j++) {} // busywait
+    }
+
+    exit(0);
```
kkoser, line 16: Same question

src/cmd_line.c
```diff
@@ -50,10 +50,15 @@ void cmd_line_attempt(const char *line) {
...
+    } else if (strcmp("memory_demo", first_word) == 0) { // temporary, for debugging
         uint32_t identifier = permissions_capability_create();
-        run("/BIN/TEST_USE.NUN", identifier);
-        run("/BIN/TEST_USE.NUN", identifier);
+
+        // set the memory
+        int difference = get_max_memory() - get_current_memory_usage();
+        capability_set_max_memory(identifier, difference);
+        run("/BIN/PRINT_EV.NUN", identifier);
+        console_printf("Attempting to run a process I don't have memory for:\n");
```
kkoser: If this is expected to error, can you make it clear with this print that an error is supposed to happen?

src/bin/print_even.c
zachwaterson, line 16: We return because main demands a return value, we exit because I don't know how to get exit code to run automatically after a process exits yet.

kkoser:
Tested and confirmed

src/bin/print_even.c
shuyangli, line 16: Yea for exit code to run automatically we need to do something special with our linker, same as the `_start` function

src/process.h
```diff
@@ -43,12 +43,12 @@ struct process {
     uint32_t entry;
     struct fd fd_table[PROCESS_MAX_OPEN_FILES];
     struct process *parent;
-    struct list children;
```
shuyangli, line 46: have we decided on this parent-children problem? when we keep a children list, when we add the child process to this list, we will add it to the ready queue, which is a side effect of the list module. it's clearly wrong, but there's no easy way to fix it without reworking the list module.

kkoser, line 46: For now, we are going to let the children live. We will tackle this issue later

kkoser, line 46: There are several easy fixes: wrapping the child list objects in a struct with a pointer to the child, or using a process table and passing around PIDs instead of process pointers.

src/syscall_handler_memory.c
```diff
@@ -0,0 +1,39 @@
...
+int32_t sys_capability_set_max_memory(uint32_t identifier, uint32_t pages) {
+    // check if identifier is owned by the process
+    if (!capability_owned_by_process(identifier, current)) {
+        return -1;
+    }
+
+    // check if page request is valid given current usage
+    int pages_left = sys_max_memory() - sys_current_memory_usage();
+    if (pages > pages_left) {
+        return -1;
```
alexmwu, line 28: do we want different error codes for identifier owned, valid capability request, and no capability (when it returns -1 below)?

zachwaterson, line 28: Eventually, but we haven't established any logic for error codes right now, so -1 is a placeholder for them all

shuyangli:
looks good

PR #70, opened 2016-05-04
Remove size limitation of user processes

zachwaterson:
This reworks run()'s use of kmalloc such that the size limit of kmalloc is no longer also the upper limit of the size of user processes. Now, run() copies the process data in chunks, as opposed to all at once.


shuyangli:
Description isn't quite accurate - It still uses kmalloc, just doesn't rely on the actual file size

shuyangli:
looks good

kkoser:
verified +1

PR #71, opened 2016-05-04
Adjust process_create and reuse pagetable_getmap in run()

zachwaterson:
It was discovered that we needed to pass in the size of the code to process_create() in order for it to allocate enough pages for the process data to be stored at the entry point.

In the same vein, we need to pagetable_getmap() for each chunk of process we move over, since we might be going into a new page.

### DIFF
--- a/src/syscall_handler_process.c
+++ b/src/syscall_handler_process.c
@@ -10,7 +10,7 @@ See the file LICENSE for details.
 #include "memorylayout.h" // PROCESS_ENTRY_POINT
 #include "permissions_capabilities.h"
 
-#define PROCESS_COPY_CHUNK 3500
+#define PROCESS_COPY_CHUNK 2048 // Half a page
 
 int32_t sys_exit(uint32_t code) {
     process_exit((int32_t)code);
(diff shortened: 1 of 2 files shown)


src/syscall_handler_process.c
```diff
@@ -10,7 +10,7 @@ See the file LICENSE for details.
 #include "memorylayout.h" // PROCESS_ENTRY_POINT
 #include "permissions_capabilities.h"
 
-#define PROCESS_COPY_CHUNK 3500
+#define PROCESS_COPY_CHUNK 2048 // Half a page
```
kkoser: can we use PAGE_SIZE / 2 here?

zachwaterson: Good thought; I didn't realize PAGE_SIZE was in the namespace of the handler. Updating now

alexmwu:
why page_size/2 just out of curiosity?

kkoser:
It needs to be a number that will go evenly into PAGE_SIZE so that we don't have to deal with stopping copy chunks that would cross pages.

alexmwu:
ah and < kmalloc size got it.

alexmwu:
LGTM

PR #72, opened 2016-05-05
Fs Syscalls for Permissions

rboccabe:
In addition to adding FS Syscalls for permission capabilities modification, this gives the first dummy process limits to the files ystem (root permissions, instead of an empty list). It provides a demo which can be tested from the console by typing fs_test. It also fixes a kmalloc error message, 

Unfortunately things are looking bleak for this to be worked in without a Guru issue, but it was tested by commented out code irrelevant to what was being tested (console_printfs during bootup and in fs_terminal_commands).

Also unfortunately, the memory_demo must be run correctly to prevent a bad machine state, but the code for the demo itself, once it's running on a good setup, works nicely.

### DIFF
--- a/src/sys_permissions.h
+++ b/src/sys_permissions.h
@@ -36,4 +36,12 @@ static inline void permissions_capability_delete(uint32_t identifier) {
     syscall(SYSCALL_capability_delete, identifier, 0, 0, 0, 0);
 }
 
+static inline int32_t permissions_capability_add_allowance(uint32_t identifier, const char *path, bool do_allow_below) {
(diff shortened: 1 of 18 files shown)


src/sys_permissions.h
```diff
@@ -36,4 +36,12 @@ static inline void permissions_capability_delete(uint32_t identifier) {
     syscall(SYSCALL_capability_delete, identifier, 0, 0, 0, 0);
 }
 
+static inline int32_t permissions_capability_add_allowance(uint32_t identifier, const char *path, bool do_allow_below) {
```
alexmwu, line 50: docs please

rboccabe, line 50: done

Issue #74, opened 2016-05-20
Crash on Bootup in Virtualbox


kkoser:
The OS can crash on boot up in Virtualbox with a "Guru Error". This error appears to become more frequent as the amount of kernel code increases. This can be tested by adding extra code such as print statements anywhere in the kernel. 

After investigation, we believe this may be due to a corruption of the interrupt table by kernel code.

Shashwatsh:
add files.iso too

kkoser:
After investigation, this is caused by the large mouse_draw_buffer static allocation in mouse.h. This buffer is used to store previous screen state before mouse overdraw, and is larger than a single page and therefore cannot be kmalloc'd

kkoser:
Fixed by giving mouse draw buffer two direct pages in memory. Should revisit this later when allocations are more well-rounded

dthain:
Nice catch!

PR #75, opened 2017-02-12
Fix guru bug, add crossplatform build support

kkoser:
The guru bug was caused by a large static allocation in the mouse draw buffer. This caused kernel space to extend into the interrupt table and corrupt it, causing triple faults. To fix this (for now) we alloce space for this during mouse_init. This buffer size (30x30x8) is too large to be allocated with kmalloc, which does not allow for allocations across pages currently. So we use the direct memory_alloc_page instead. We call this method twice in order to reserve two adjacent pages. The adjacency of the allocations is not guranteed, but can be implied from the early level in which this code happens. Once we can allocate memory larger than a page safely, we should revisit this issue

    Also add the ability to build nunya on multiple filesystems by abstracting away commands used for building. You can override $CC, $LD, and $ISOGEN to define your own paths to the tools needed to build. Default values have been provided to ensure backwards compatability with older setups.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -22,42 +22,45 @@ LIB_INCLUDE_PATH = ./include
 KERNEL_CCFLAGS = -Wall -c -ffreestanding -m32 -march=i386
 KERNEL_LDFLAGS = -m elf_i386
 
+LD?=ld
+CC?=gcc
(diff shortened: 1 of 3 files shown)


kkoser:
Fixes #74

